PR #159, opened 2026-04-18
fix: correct publish_lock scope for Maven and Cargo

devitway:
## Summary

- **Maven**: lock guard was scoped inside `if` block, dropping before `storage.put()` and `update_artifact_metadata()`. Concurrent uploads of different versions could produce incomplete `maven-metadata.xml`. Now locks on metadata key covering the entire upload+metadata cycle.
- **Cargo**: lock key was per-version (`crate_key`) but the read-modify-write target is the per-crate sparse index file. Concurrent publishes of different versions of the same crate could lose index entries. Now locks on `index_key` matching the actual RMW target.
- Add `scripts/lock-audit.sh` for CI constraint checking of lock patterns.
- Add `Makefile` with `make check` (fmt, clippy, test, coherence, lock-audit).

## Test plan

- [x] `cargo check` passes
- [x] 596 existing tests pass (no regressions)
- [x] `scripts/lock-audit.sh` reports no FAIL on patched code
- [ ] Manual: concurrent Maven deploy of two versions, verify both appear in metadata.xml
- [ ] Manual: concurrent Cargo publish of two versions, verify both in sparse index

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,40 @@
+# NORA — build & quality pipeline
+# Usage: make check   — run all quality checks
+#        make test    — unit tests only
+#        make build   — release build
+#        make release — tagged release (runs checks first)
+
+CARGO := cargo
+
+.PHONY: check test build release fmt clippy coherence lock-audit
+
+check: fmt clippy test coherence lock-audit
+	@echo ""
+	@echo "=== All checks passed ==="
+
+fmt:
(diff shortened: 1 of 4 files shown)


devitway:
Superseded by #160 which includes all changes from this PR plus flaky test fixes and Docker pull counter lock.

PR #191, opened 2026-04-26
feat: Add pub repo

### LINKED ISSUE #166: Add Pub proxy support (Dart packages)
## Summary

Add proxy support for pub.dev to cache and serve Dart/Flutter packages from the upstream pub repository.

## Motivation

- **CI acceleration**: Dart/Flutter builds run `pub get` on every run. A local proxy caches `.tar.gz` archives and metadata, reducing network latency and cost.
- **Air-gapped / restricted networks**: Projects in finance, government, or on-premise environments cannot access pub.dev directly. A local proxy provides a compliant package source.
- **Resilience**: Outages of pub.dev or its CDN break builds. A NORA proxy acts as a transparent mirror, serving previously cached packages even when upstream is down.
- **Underserved niche**: While some solutions exist (e.g., `unpub`, private Pub servers), none offer a simple, zero‑configuration **caching proxy** focused purely on upstream acceleration and reliability.

## Pub API (proxy scope)

Pub interacts with a repository according to the [Pub Repository Specification v2](https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md). The proxy needs to handle:

1. **Package listing**: `GET /api/packages/{package}` — returns metadata with available versions.
2. **Version metadata**: `GET /api/packages/{package}/versions/{version}` — includes the download URL and dependencies.
3. **Archive download**: `GET /packages/{package}/versions/{version}.tar.gz` — the actual package content.
4. **Search**: `GET /api/packages?q={term}` — package search
5. **Security advisories**: `GET /api/packages/<package>/advisories` — returns list security advisories for a package

The proxy intercepts requests, fetches from `https://pub.dev` if not cached, stores the `.tar.gz` files and API responses locally, and serves subsequent requests from the cache.

## Scope

**Phase 1 (this issue):** Proxy mode — cache upstream packages and metadata, serve locally.
**Phase 2 (future):** Hosted mode — publish private Dart/Flutter packages to a local repository.

mit-73:
## What

Add proxy support for pub.dev to cache and serve Dart/Flutter packages from the upstream pub repository.

## Why

Closes https://github.com/getnora-io/nora/issues/166

## Checklist

- [x] New registry? See CONTRIBUTING.md checklist


nora-registry/src/metrics.rs
```diff
@@ -119,6 +119,10 @@ fn detect_registry(path: &str) -> String {
         "npm".to_string()
     } else if path.starts_with("/cargo") {
         "cargo".to_string()
+    } else if path.starts_with("/api/packages")
+        || (path.starts_with("/packages/") && path.contains("/versions/"))
```
mit-73, line 123: Collecting metrics via API paths is not very reliable; at least for now, there may be problems with `pypi` due to the `/packages` path.

It seems like the `detect_registry` method should rely on something else (for example, headers) or do its job differently?

nora-registry/src/registry/pub_dart.rs
```diff
@@ -0,0 +1,975 @@
+// Copyright (c) 2026 Volkov Pavel | DevITWay
```
mit-73: What Copyright should I write here?

mit-73:
@devitway, unfortunately, I wasn't able to build a release version and fully test everything due to the OS. The project currently requires builds to be run strictly on Unix systems. I tested everything I could, and it seems to work.

I'll also leave a couple of code points that would be better fixed:
- the architecture doesn't isolate code chunks
- the first point makes cohesion suffer (to add a new registry, I had to edit 10 files), which creates more opportunities for bugs
- DRY, it's largely ignored here, which leads to similar code chunks
- I would separate the UI into a separate module (completely in TS + react/vue/svelte/...). This will allow it to be developed separately and more freely
- There are no instructions for the initial installation of the environment (maybe I didn’t find it myself)

nora-registry/src/registry/pub_dart.rs
devitway: Use `// Copyright (c) 2026 The NORA Authors` and `// SPDX-License-Identifier: MIT`.

We are standardizing all headers to "The NORA Authors" across the codebase.

mit-73: done

devitway:
Incorporated into v0.7.0 with changes (curation wiring, dynamic registry loading, URL rewrite). Credited as co-author in the release commit. Thanks!

PR #232, opened 2026-05-02
chore: remove stale advisory ignore from deny.toml

devitway:
## Summary
- Remove `RUSTSEC-2025-0119` (number_prefix/indicatif) from deny.toml — dependency no longer in tree

## Test plan
- [x] `cargo deny check` → advisories ok, bans ok, licenses ok, sources ok


devitway:
Superseded by #233 which included this deny.toml change in the same merge.

PR #254, opened 2026-05-04
fix(s3): add colon to uri_encode whitelist for sha256 digest paths

Kzazagzo:
## What
Add `:` to `uri_encode` whitelist in S3 storage backend.

## Why
Docker image blobs are addressed as `sha256:<digest>`, where `:` is a valid
path character per RFC 3986. The previous implementation percent-encoded `:`
to `%3A` in the canonical URI, while reqwest sent the raw colon in the actual
HTTP request. Garage S3 rejected all blob GET/PUT/HEAD with `403 Invalid
signature` due to this mismatch.

## Checklist
- [ ] Updated CHANGELOG.md (if user-facing change)

### DIFF
--- a/nora-registry/src/storage/s3.rs
+++ b/nora-registry/src/storage/s3.rs
@@ -258,12 +258,12 @@ impl S3Storage {
     }
 }
 
-/// URL-encode a string for S3 canonical URI (encode all except A-Za-z0-9-_.~/)
+/// URL-encode a string for S3 canonical URI (encode all except A-Za-z0-9-_.~/:)
 fn uri_encode(s: &str) -> String {
     let mut result = String::with_capacity(s.len() * 3);
     for c in s.chars() {
         match c {
-            'A'..='Z' | 'a'..='z' | '0'..='9' | '-' | '_' | '.' | '~' | '/' => result.push(c),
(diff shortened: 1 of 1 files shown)


devitway (approved):
LGTM. The fix is correct.

**Root cause**: `reqwest` sends raw `:` in URL paths (valid per RFC 3986), but `uri_encode()` encoded it to `%3A` in the canonical URI for SigV4 signing. Strict S3 implementations (Garage) compute the signature from the received URL (raw `:`), causing a mismatch → 403.

**Why we missed it**: MinIO and RustFS normalize the canonical URI before signature verification, so the mismatch is invisible.

One suggestion for follow-up: we should add a unit test for this case:
```rust
#[test]
fn test_uri_encode_colon_for_docker_digests() {
    // Colon must not be encoded — reqwest sends raw ':' in URLs
    assert_eq!(uri_encode("sha256:abc123"), "sha256:abc123");
    assert_eq!(uri_encode("docker/sha256:abc/blobs/sha256:def"), "docker/sha256:abc/blobs/sha256:def");
}
```

I'll add this test on our side. Welcome to the project, @Kzazagzo!

PR #265, opened 2026-05-07
ci(release): add binary smoke test before publish

### LINKED ISSUE #264: The 0.8.1 binary file failed to build.
### Description

The 0.8.1 binary file failed to build.

### Expected Behavior

_No response_

### Steps to Reproduce

·

### NORA Version

v0.8.1

### Registry Protocol

None

### Storage Backend

None

### Relevant Logs

```shell

```

devitway:
## Summary

- Add binary smoke test in `build` job — verifies the extracted binary is a valid ELF, executes `--version`, and responds on `/health` endpoint before Docker images are pushed
- Add binary validation in `release` job — lightweight ELF + `--version` check after downloading artifact, before signing and publishing to GitHub Releases

## Problem

The release pipeline smoke-tested only the Docker image (`/health` on Alpine container) but never verified that the standalone `nora-linux-amd64` binary actually works. A broken binary could be published to GitHub Releases without detection (#264).

## What changed

**`.github/workflows/release.yml`** — two new steps:

| Job | Step | Checks |
|-----|------|--------|
| `build` | Smoke test binary | ELF 64-bit validation, static linking warning, `--version`, `/health` endpoint |
| `release` | Validate binary artifact | ELF 64-bit validation, `--version` |

The build job test runs **before** Docker image push — if the binary is broken, the pipeline fails early without publishing anything.

## Test plan

- [ ] Trigger release with a valid tag — pipeline passes, binary published
- [ ] Simulate broken binary (e.g. truncated file) — pipeline fails at smoke test step before any images are pushed

Closes #264

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,6 +55,21 @@ jobs:
           echo "Binary size: $(du -sh nora-linux-amd64 | cut -f1)"
           file ./nora-linux-amd64
 
+      - name: Smoke test binary
+        run: |
+          echo "--- Verifying binary is a valid static executable ---"
+          file ./nora-linux-amd64 | grep -q "ELF 64-bit" || { echo "::error::Binary is not a valid ELF 64-bit executable"; exit 1; }
+          ldd ./nora-linux-amd64 2>&1 | grep -q "not a dynamic executable\|statically linked" || echo "::warning::Binary is dynamically linked"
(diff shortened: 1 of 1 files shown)


devitway:
Superseded by #269 which addresses the root cause — Trivy scan timeout blocking the release job. Binary smoke test is no longer needed as a separate fix.

PR #347, opened 2026-05-15
refactor: extract build_manifest_response helper (#338)

### LINKED ISSUE #338: refactor: extract manifest response builder in get_manifest
## Context

`get_manifest` in `docker.rs` has three return paths that build manifest responses with identical header sets (`Content-Type`, `Docker-Content-Digest`, `Content-Length`). The third path (library/ fallback) was missing `Content-Length` until #337 — the duplication made it easy to miss.

## Proposal

Extract a helper function like `build_manifest_response(content_type, digest, data) -> Response` to eliminate the triple copy-paste. This will prevent future header drift between the paths.

## Scope

- `nora-registry/src/registry/docker.rs` — `get_manifest` function
- No behavior change, pure refactor

## Milestone

v0.9.0

devitway:
## Summary

- Extract `build_manifest_response` helper from `get_manifest` to eliminate triple copy-paste of response headers (`Content-Type`, `Docker-Content-Digest`, `Content-Length`)
- Returns `http::response::Builder` (not final `Response`) so call sites can append headers before `.body()` — extensible by design
- Replace `.expect()` with `match` + 500 fallback to avoid panics on malformed header values
- Add `Content-Type`, `Content-Length`, `Docker-Content-Digest` header assertions to `test_docker_put_get_manifest`

Closes #338

## Motivation

The header duplication caused #337 (missing `Content-Length` in library/ fallback path). This refactor makes such drift structurally impossible — headers are defined in exactly one place.

## Test plan

- [x] `cargo fmt --check` — pass
- [x] `cargo clippy -D warnings` — pass
- [x] 93 docker tests pass (including new header assertions)
- [x] Negative security tests: 10 pass
- [x] API contract tests: 23 pass
- [x] OCI conformance: 12 pass

### DIFF
--- a/nora-registry/src/registry/docker.rs
+++ b/nora-registry/src/registry/docker.rs
@@ -937,17 +937,15 @@ async fn get_manifest(
             meta_key,
         ));
 
-        let content_length = data.len().to_string();
-        return (
-            StatusCode::OK,
-            [
-                (header::CONTENT_TYPE, content_type),
-                (HeaderName::from_static("docker-content-digest"), digest),
-                (header::CONTENT_LENGTH, content_length),
-            ],
-            data,
-        )
-            .into_response();
+        return match build_manifest_response(&content_type, &digest, data.len())
+            .body(axum::body::Body::from(data))
+        {
+            Ok(resp) => resp.into_response(),
+            Err(e) => {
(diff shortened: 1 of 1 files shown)


devitway:
Superseded by #349 which includes this refactor as part of the v0.9.0 milestone.

PR #351, opened 2026-05-16
fix(raw): replace hardcoded immutable Cache-Control with configurable default

devitway:
## Summary

- The raw registry previously hardcoded `Cache-Control: public, max-age=31536000, immutable` on all downloads and HEAD responses
- This is incorrect for mutable raw artifacts (overwrite-on-PUT is allowed) — stale cached content breaks CI pipelines
- Now configurable via `NORA_RAW_CACHE_CONTROL` env var (or `raw.cache_control` in TOML), defaulting to `no-cache`

## Changes

- `config.rs`: added `cache_control` field to `RawConfig` with `no-cache` default + env override
- `registry/raw.rs`: replaced two hardcoded header values with config reference
- `test_helpers.rs`: added field to test context builder
- Added integration test verifying the default header value


devitway:
Content already on main (merged via history rewrite)

PR #353, opened 2026-05-17
feat: terraform upstream URL resolution and UI prerelease filtering

devitway:
## Summary
- Terraform provider binary downloads now resolve the real upstream URL (releases.hashicorp.com) from cached metadata instead of incorrectly proxying through registry.terraform.io
- Original download URL preserved in `_nora_upstream_url` metadata field during rewrite
- UI detail views for all registries support `?prerelease=true` and `?all=true` query params
- Pre-release detection covers semver, PyPI PEP 440, and RubyGems conventions
- Default view shows top 20 stable versions with prerelease toggle

## Test plan
- [x] `cargo test --lib --bin nora` — 974 tests pass
- [x] `cargo clippy -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] Semgrep — no new findings
- [x] Pre-commit version check — pass
- [ ] CI status checks


devitway:
Prerelease filtering merged via metadata feature. Terraform URL part moved to #355

PR #457, opened 2026-05-24
ci: remove Astra from Trivy scan matrix

devitway:
## Summary

- Astra base image consistently fails Trivy scan — Java DB download (870 MB) fills disk on self-hosted runner
- Astra registry is not indexed by Trivy, scan provides no value

## Test plan

- [ ] Scan matrix runs only alpine + redos


devitway:
Astra scan failure is transient (disk full on runner, not Trivy incompatibility). Astra scanned successfully on v0.9.0 and v0.9.1.

PR #716, opened 2026-06-14
fix(config): allow partial config.toml by adding serde defaults for server and storage

heunghingwan:
When config.toml exists but lacks certain values (e.g. host / port in [server], or a [server]/[storage] section entirely), serde deserialization failed and the app could not start. With these changes all unset fields fall back to sensible defaults.

- Added `#[serde(default)]` to `server` and `storage` in `Config`
- Added `#[serde(default = "...")]` for `host` and `port` in `ServerConfig`
- Added `impl Default` for `ServerConfig` and `StorageConfig`
- Updated CHANGELOG and added test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 ## [Unreleased]
 
+### Fixed
+- **Partial `config.toml`** — missing `[server]`, `[storage]`, or fields like `host`/`port` no longer prevent startup; serde defaults are applied for all unset values.
+
 ## [0.9.4] - 2026-06-13
 
 ### Added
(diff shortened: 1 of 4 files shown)


devitway (approved):

devitway (approved):
LGTM — clean, correctly scoped fix.

Verified locally: `cargo fmt --check`, `cargo clippy -p nora-registry --lib`, and `cargo test -p nora-registry config::` all pass (132 tests, including the new `test_config_from_toml_partial`).

`server` and `storage` were the only two top-level `Config` fields without `#[serde(default)]` — the other 24 already had it — so this completes the set: an empty or partial `config.toml` now deserializes and passes validation. The added default values match the existing `Config::default()` field-for-field, so there's no behavioral drift.

Thanks for the contribution! Filed #717 as a small follow-up to collapse the now-redundant inline literals in `Config::default()` onto `ServerConfig::default()` / `StorageConfig::default()` — that's pre-existing and not something this PR needs to touch.

PR #720, opened 2026-06-14
feat: add npm /-/whoami endpoint for token-based identity

heunghingwan:
## Summary

Implement the npm `/-/whoami` endpoint so that `npm whoami` works with tokens and basic auth.

### `nora-registry/src/auth/mod.rs`

- Add `AuthenticatedUser(String)` struct that carries the authenticated username through request extensions
- Insert it in all auth paths (token, OIDC, basic auth, anonymous, auth-disabled)
- Exclude `/npm/-/whoami` from `anonymous_read` shortcut — otherwise it would always return `"anonymous"` regardless of credentials

### `nora-registry/src/registry/npm.rs`

- Add `handle_whoami()` that returns `{"username": "..."}` for `GET /npm/-/whoami`
- Wire it into the existing wildcard handler before any package lookup logic

## Why

npm CLI calls `GET /-/whoami` when resolving credentials for token-based auth (see `@npmcli/get-identity.js`). Without this endpoint, `npm whoami` and any command that internally verifies identity with a token will fail with `ENEEDAUTH`.

## How to use

Create an API token from the dashboard UI (`/ui/tokens`), then configure `.npmrc`:

```
//your-registry/npm/:_authToken=nra_xxxxxxxxxxxx
//your-registry/:_authToken=nra_xxxxxxxxxxxx
```

Due to npm's `nerfDart()` URL normalization inconsistency between `@npmcli/config` and `npm-registry-fetch`, two lines are needed to cover both `npm whoami` and `npm publish`.

## Limitations

`npm login` / `npm adduser` are **not** supported. Nora does not implement the CouchDB-style endpoints (`/-/v1/login`, `/-/user/org.couchdb.user:*`). To support those, the following would need to be added:

- `POST /npm/-/v1/login` — web-based OAuth login flow
- `PUT /npm/-/user/org.couchdb.user:{username}` — legacy CouchDB login
- Session token handling and OTP support

## Testing

- `test_npm_whoami_anonymous` — auth disabled, returns `{"username":"anonymous"}`
- `test_npm_whoami_authenticated` — basic auth, returns the authenticated username
- `test_npm_whoami_requires_auth` — no credentials, returns 401
- All 1353 existing tests pass

### DIFF
--- a/nora-registry/src/auth/mod.rs
+++ b/nora-registry/src/auth/mod.rs
@@ -18,6 +18,10 @@ pub use namespace::{enforce_namespace_scope, NamespaceAuthority};
 pub use oidc::OidcValidator;
 pub use token_routes::{token_routes, TokenListItem, TokenListResponse};
 
+/// Authenticated username carried in request extensions after successful auth.
+#[derive(Clone, Debug)]
+pub struct AuthenticatedUser(pub String);
+
 use axum::{
     body::Body,
     extract::{ConnectInfo, State},
@@ -177,6 +181,9 @@ pub async fn auth_middleware(
         request
             .extensions_mut()
             .insert(NamespaceAuthority::Unrestricted);
+        request
+            .extensions_mut()
+            .insert(AuthenticatedUser("anonymous".to_string()));
         return next.run(request).await;
     }
 
@@ -185,6 +192,9 @@ pub async fn auth_middleware(
         request
             .extensions_mut()
             .insert(NamespaceAuthority::Unrestricted);
+        request
+            .extensions_mut()
+            .insert(AuthenticatedUser("anonymous".to_string()));
         return next.run(request).await;
     }
 
@@ -199,7 +209,10 @@ pub async fn auth_middleware(
     // Token management always requires auth, even with anonymous_read
     let is_token_management = path.starts_with("/ui/tokens") || path.starts_with("/api/ui/tokens");
 
-    // Allow anonymous read if configured (but not for Docker /v2/ or token management)
+    // npm whoami always requires auth (otherwise it returns "anonymous" for every user)
+    let is_whoami = path.ends_with("/-/whoami");
+
+    // Allow anonymous read if configured (but not for Docker /v2/, token management, or whoami)
     let is_read_method = matches!(
         *request.method(),
         axum::http::Method::GET | axum::http::Method::HEAD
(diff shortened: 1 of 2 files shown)


devitway (approved):
Reviewed the auth trust boundary carefully — no bypass:

- Every success path in `auth_middleware` inserts `AuthenticatedUser` (token, OIDC, basic, plus the three anonymous/auth-disabled/anonymous-read paths), so the npm handler's `Extension<AuthenticatedUser>` extractor can never 500 on a missing extension.
- `/-/whoami` is correctly excluded from the `anonymous_read` shortcut — that makes it *more* restrictive (requires real auth), not a bypass — and `is_public_path` does not match `/npm/*`, so it can't leak an anonymous identity.
- Behaviour: auth disabled → `"anonymous"`, valid creds → real user, auth enabled + no creds → 401. Matches the three new tests.

Verified locally: build + `clippy -D warnings` clean, the 3 whoami tests pass, 90 npm tests green.

One follow-up I'll take directly (not blocking): `handle_whoami` builds the body with `format!(r#"{{"username":"{}"}}"#, ...)`, which doesn't escape the username — a name containing `"` (reachable via the OIDC `sub` claim, which isn't charset-validated) injects into the JSON. I'll fix it with `serde_json` in a follow-up patch.

Merging now — thanks for the clean, well-tested contribution!

heunghingwan:
Followed up on the "two lines needed" point in the PR description after tracing the actual npm CLI code more carefully.

The root cause isn't really a nerfDart() inconsistency — it's that `npm whoami` and `npm publish` derive the registry URL from different sources:

**`npm whoami`**
`flatOptions.registry` = the raw `registry` config value (e.g. `https://registry.example.com/npm/` with trailing slash).

**`npm publish` (for a scoped package like `@mycorp/foo`)**
`npmFetch.pickRegistry(resolved, opts)` → checks `opts['@mycorp:registry']` first, which may be `https://registry.example.com/npm` (no trailing slash, depending on how the user wrote it).

So both `//registry.example.com/npm/:_authToken=...` and `//registry.example.com/npm:_authToken=...` are needed because two different code paths produce two different URL strings, and each matches a different nerfDart key. The nerfDart normalization fix ensures trailing-slash differences don't cause a mismatch, which is correct — but the real source of the divergence is that `pickRegistry` returns the scope-specific registry while `flatOptions.registry` returns the default registry, and the user may configure them with slightly different formats.

Not a bug in this PR at all — just wanted to document the finding for anyone wondering why two lines are necessary.

PR #782, opened 2026-06-24
fix(cargo): fix sparse index api base for cargo publish

tomchon:
## Summary
This fixes the Cargo sparse index `config.json` response so the advertised `api` base matches Cargo's registry protocol expectations.

## Problem
`/cargo/index/config.json` currently returns an `api` value ending in `/cargo/api`.

Cargo appends `/api/v1/...` to the advertised `api` base for registry web API requests, so the current value causes publish requests to be sent to:

`/cargo/api/api/v1/crates/new`

That path does not exist in NORA. The actual publish route is:

`/cargo/api/v1/crates/new`

As a result, `cargo publish` receives a `404 Not Found` even though the Cargo publish handler is implemented.

## Fix
Return the registry mount as the `api` base:

- before: `{base}/cargo/api`
- after: `{base}/cargo`

This makes Cargo construct the expected routes for metadata and publish requests.

## Test Update
The existing assertion for `config.json` was updated to reflect Cargo's expected `api` base.

## Reproduction
Observed with a real Cargo client while debugging a failed `cargo publish` against a NORA instance.
The client was issuing:

`PUT /cargo/api/api/v1/crates/new`

After tracing that back to the `config.json` `api` value, this patch updates the advertised base accordingly.

### DIFF
--- a/nora-registry/src/registry/cargo_registry.rs
+++ b/nora-registry/src/registry/cargo_registry.rs
@@ -1170,7 +1170,7 @@ mod integration_tests {
 
     /// Verify that config.json `api` field produces correct metadata route (#442).
     ///
-    /// Cargo constructs `{api}/v1/crates/{name}` for metadata requests.
+    /// Cargo constructs `{api}/api/v1/crates/{name}` for metadata requests.
     /// The `api` field must match the route prefix so requests don't 404.
     #[tokio::test]
     async fn test_cargo_config_api_routes_to_metadata() {
@@ -1190,9 +1190,9 @@ mod integration_tests {
         let config: serde_json::Value = serde_json::from_slice(&config_body).unwrap();
         let api_url = config["api"].as_str().unwrap();
 
-        // Construct the metadata URL the way Cargo does: {api}/v1/crates/{name}
+        // Construct the metadata URL the way Cargo does: {api}/api/v1/crates/{name}
         let metadata_path = format!(
-            "{}/v1/crates/serde",
+            "{}/api/v1/crates/serde",
             api_url.trim_start_matches("http://localhost")
         );
 
--- a/nora-registry/src/registry/cargo_registry.rs
+++ b/nora-registry/src/registry/cargo_registry.rs
@@ -56,7 +56,9 @@ async fn index_config(State(state): State<AppState>) -> Response {
     let base = nora_base_url(&state);
     let config = serde_json::json!({
         "dl": format!("{}/cargo/api/v1/crates", base),
-        "api": format!("{}/cargo/api", base)
+        // Cargo appends `/api/v1/...` to this base for write and metadata requests,
+        // so the advertised API root must be the registry mount, not `/cargo/api`.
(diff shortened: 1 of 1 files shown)


tomchon:
One detail that helped confirm the root cause during debugging:

A manual request to the implemented publish route, `PUT /cargo/api/v1/crates/new`, does reach the Cargo publish handler and returns a payload parsing error as expected when the body is not in Cargo's wire format.

However, a real `cargo publish` request was observed going to:

`PUT /cargo/api/api/v1/crates/new`

So this is not a missing publish handler. The extra `/api` comes from the `api` value advertised in `/cargo/index/config.json`, because Cargo appends `/api/v1/...` to that base when constructing registry web API requests.

ertime037 (dismissed):

devitway (dismissed):

devitway:
Thanks for tracking this down — the `api` base fix is correct.

CI is failing on one remaining test that still carries the previous assumption: `test_cargo_config_api_routes_to_metadata`. It builds the request URL as `{api}/v1/crates/{name}`, but Cargo appends the full `/api/v1/...` to the advertised `api` base. With the corrected `api = {base}/cargo`, the request should be `{api}/api/v1/crates/serde` (i.e. `/cargo/api/v1/crates/serde`), which matches the metadata route.

Could you update that test (and its doc/inline comments) in the same PR? The change is:

```rust
// before
"{}/v1/crates/serde",
// after
"{}/api/v1/crates/serde",
```

and the comments from `{api}/v1/crates/{name}` to `{api}/api/v1/crates/{name}`. Same root cause as the `config.json` change — that test just encoded the old base, so flipping the suffix lines it up with the corrected `api` value.

tomchon:
Updated the metadata-route test and its inline comments to match Cargo's `api` base semantics.

The test now constructs the metadata request as `{api}/api/v1/crates/{name}` (instead of `{api}/v1/...`), which lines up with the corrected `api = {base}/cargo` value from `config.json`.

ertime037 (dismissed):

tomchon:
Updated per your suggestion.

In addition to fixing the existing metadata-route test/comments to use `{api}/api/v1/crates/{name}`, I also added a publish-route test derived from the advertised `api` base.

That new test constructs the publish path as `{api}/api/v1/crates/new` and verifies it reaches the publish handler successfully. My goal there was to lock in the `config.json` `api` semantics for both metadata and publish paths, so we don't regress back to `/cargo/api/api/...` later.

ertime037 (approved):

devitway:
Landed in #783 (merged to `main` as 1380f34) — thank you for the fix and the careful root-cause work distinguishing the publish handler from the advertised `api` base.

`main` requires signed commits and merges through a merge queue, which blocked merging this branch directly (the commits here aren't signed). To get it in without losing your work, I relanded the exact change as a signed commit in #783, with you credited as co-author. Both tests you added (metadata and publish routes derived from the `api` base) came along unchanged.

Closing in favor of #783. Much appreciated!

PR #789, opened 2026-06-27
Potential fixes for 2 code quality findings

ertime037:
_This PR applies 2/2 suggestions from code quality [AI findings](https://github.com/getnora-io/nora/security/quality/ai-findings)._

### DIFF
--- a/docs-site/src/content/docs/observability/prometheus-metrics.md
+++ b/docs-site/src/content/docs/observability/prometheus-metrics.md
@@ -114,4 +114,4 @@ scrape_configs:
 
 - [Settings](/configuration/settings/) — server configuration reference (the `/metrics` endpoint is always enabled)
 - [Circuit Breaker](/configuration/circuit-breaker/) — breaker metrics details
-- [Upgrade Guide](/deployment/upgrade-guide/) — new metrics in v0.9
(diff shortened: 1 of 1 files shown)


devitway:
Thanks — the table did mix `*` and `/*` notation, so flagging it is fair.

Closing this one though: the autofix standardizes everything to `/go*`, `/raw*`, … which mis-documents the actual behavior. The `registry` label comes from `detect_registry()` in `nora-registry/src/metrics.rs`, which matches a **literal path prefix** via `starts_with`. The trailing slash is load-bearing for go/raw/gems/terraform/ansible/nuget/pub/conan — `path.starts_with("/go/")`, not `"/go"`. The unit tests pin this: `detect_registry("/goblin/...") == "other"` and `detect_registry("/rawdata/...") == "other"`. So `/go*` would tell readers that `/goblin` maps to `go`, which is wrong.

The mixed notation reflects real, intentional asymmetry in the code (docker/maven/npm/cargo/pypi/ui match a bare prefix; the rest require the trailing slash), so the right fix is to remove the glob ambiguity rather than homogenize it. Documenting the exact `starts_with` prefixes (trailing slash shown, first-match-wins noted) separately; the Upgrade Guide caption tweak is good and folded into that change.

PR #795, opened 2026-07-02
feat(storage): add s3_virtual_hosted toggle for virtual-hosted-style requests

synchrone:
## Problem

The S3 backend hardcodes path-style addressing (`with_virtual_hosted_style_request(false)` in `storage/s3.rs`). Some providers reject **signed** path-style requests entirely — Alibaba Cloud OSS answers them with `403 SecondLevelDomainForbidden` ("Please use virtual hosted style to access") on both its native `oss-<region>.aliyuncs.com` and S3-compat `s3.oss-<region>.aliyuncs.com` endpoints. Since every storage operation fails, `/health` reports 503 and NORA can't run against OSS at all.

(Subtle footgun for anyone testing this: OSS answers *unsigned* path-style requests with a generic `NoSuchBucket`, which makes path-style look supported until you sign the request.)

## Change

- New `storage.s3_virtual_hosted` config field (config.toml) with `NORA_STORAGE_S3_VIRTUAL_HOSTED` env override, following the existing `apply_env_overrides` bool pattern
- **Default `false`** — behavior is unchanged for all existing deployments (MinIO/RustFS/SeaweedFS/Garage generally want path-style)
- Threaded through `Storage::new_s3` / `S3Storage::new` into `AmazonS3Builder::with_virtual_hosted_style_request`
- Logged in the "Using S3 storage" startup line
- Docs: settings tables (EN/RU), new "Alibaba Cloud OSS" example section + troubleshooting entry in the S3 storage guide

## Verification

- Against OSS (cn-hongkong), same SigV4 credentials: path-style `ListObjectsV2` → `403 SecondLevelDomainForbidden`; virtual-hosted → `200`
- `cargo test --lib --bin nora`: 1493 passed (extends `test_env_override_storage`, adds default-off/truthy-parsing test and a vhost constructor test)
- `cargo clippy --package nora-registry --all-targets -- -D warnings` and `cargo fmt --check` clean

### DIFF
--- a/docs-site/src/content/docs/configuration/s3-storage.md
+++ b/docs-site/src/content/docs/configuration/s3-storage.md
@@ -15,7 +15,7 @@ By default NORA stores artifacts on the local filesystem. For production deploym
 | `NORA_STORAGE_S3_ACCESS_KEY` | — | Access key. If omitted, anonymous access is used |
 | `NORA_STORAGE_S3_SECRET_KEY` | — | Secret key |
 | `NORA_STORAGE_S3_REGION` | `us-east-1` | Region. Required by some S3 implementations |
-| `NORA_STORAGE_S3_VIRTUAL_HOSTED` | `false` | Use virtual-hosted-style requests (`https://<bucket>.<endpoint>/…`) instead of path-style. Required by providers that reject signed path-style requests, e.g. Alibaba Cloud OSS |
+| `NORA_STORAGE_S3_VIRTUAL_HOSTED` | `false` | Use virtual-hosted-style requests: the endpoint is used **as-is** and must include the bucket host (e.g. `https://<bucket>.oss-<region>.aliyuncs.com`). Required by providers that reject signed path-style requests, e.g. Alibaba Cloud OSS |
 
 :::caution
 NORA **does not create buckets automatically**. The bucket must exist before NORA starts. Use an init container or pre-create it manually.
@@ -183,12 +183,12 @@ Use IAM roles instead of static credentials when running on EC2/ECS/EKS. Omit `N
 
 ## Alibaba Cloud OSS
 
-OSS rejects **signed path-style requests** with `403 SecondLevelDomainForbidden` ("Please use virtual hosted style to access"), so `NORA_STORAGE_S3_VIRTUAL_HOSTED` must be enabled. Point the endpoint at the bucket's region (the bucket name is prepended to the host automatically) and use the region ID without the `oss-` prefix:
+OSS rejects **signed path-style requests** with `403 SecondLevelDomainForbidden` ("Please use virtual hosted style to access"), so `NORA_STORAGE_S3_VIRTUAL_HOSTED` must be enabled. With it enabled the endpoint is used **as-is** — it must include the bucket host. Use the region ID without the `oss-` prefix:
 
 ```yaml
 environment:
   NORA_STORAGE_MODE: s3
-  NORA_STORAGE_S3_URL: https://oss-cn-hongkong.aliyuncs.com
+  NORA_STORAGE_S3_URL: https://my-nora-registry.oss-cn-hongkong.aliyuncs.com  # bucket in the host
   NORA_STORAGE_BUCKET: my-nora-registry
   NORA_STORAGE_S3_ACCESS_KEY: LTAI...
   NORA_STORAGE_S3_SECRET_KEY: ...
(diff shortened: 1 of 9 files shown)


synchrone:
proof-of-comprehension: I reviewed this PR with my human eyes and it checks out. I need to roll a cached docker registry in China but deploying to Alibaba is blocked by this tiny assumption/hardcode.

devitway:
Thanks for this — the code change itself is clean: the toggle is threaded correctly through `StorageConfig` → `Storage::new_s3` → `S3Storage::new` into the builder, all three `new_s3` call sites in `main.rs` are updated, `#[serde(default)]` + default-`false` keeps existing deployments byte-identical, and the env-parse follows the existing `apply_env_overrides` idiom. `cargo fmt` / `clippy` / the lib tests are green locally, and the `SecondLevelDomainForbidden`-on-signed-requests footgun note is a genuinely useful catch. 🙌

I think there's one issue in the **docs**, though — and I'd like to reconcile it with your testing before merge, because as written I don't think the documented Alibaba OSS example would actually work.

### The concern

The OSS section says the bucket is prepended to the host automatically and gives a bucket-less endpoint:

```yaml
NORA_STORAGE_S3_URL: https://oss-cn-hongkong.aliyuncs.com   # no bucket
NORA_STORAGE_BUCKET: my-nora-registry
NORA_STORAGE_S3_VIRTUAL_HOSTED: "true"
```

But reading `object_store` 0.13.2, when a custom endpoint **and** `virtual_hosted_style_request = true` are both set, the endpoint is used **as-is** — the crate does *not* inject the bucket into the host:

- `aws/builder.rs:1214-1216`
  ```rust
  let bucket_endpoint = match (&endpoint, zonal_endpoint, virtual_hosted) {
      (Some(endpoint), _, true)  => endpoint.clone(),                  // used verbatim
      (Some(endpoint), _, false) => format!("{}/{}", endpoint, bucket),
      ...
  ```
- `aws/client.rs:215-216`
  ```rust
  pub(crate) fn path_url(&self, path: &Path) -> String {
      format!("{}/{}", self.bucket_endpoint, encode_path(path))
  }
  ```
- and the crate's own comment right above the match (`builder.rs:1211-1212`): *"if `virtual_hosted_style_request` is true then `endpoint` should have bucket name included."*

So with the documented config, object requests would go to `https://oss-cn-hongkong.aliyuncs.com/<key>` — with the bucket in neither the host nor the path.

For virtual-hosted to actually address the bucket, the bucket has to be in the endpoint:

```yaml
NORA_STORAGE_S3_URL: https://my-nora-registry.oss-cn-hongkong.aliyuncs.com   # bucket in host
NORA_STORAGE_BUCKET: my-nora-registry
NORA_STORAGE_S3_VIRTUAL_HOSTED: "true"
```

(The same `https://<bucket>.<endpoint>/<key>` phrasing is in the `s3.rs` / `storage.rs` doc comments too.)

### Could you share the exact endpoint you validated against?

I haven't reproduced this against a live OSS bucket, so I might be missing something — but the source paths above are pretty unambiguous, and my guess is the config you actually tested had the bucket in the host and the doc example just drifted. If you can paste the working endpoint, we can make the shipped doc match it exactly.

### Minor

`test_s3_storage_creation_virtual_hosted` only asserts `backend_name() == "s3"`, which is constant regardless of the flag — so it confirms construction doesn't panic but doesn't exercise the addressing behavior (it wouldn't have caught the above). Not a blocker, just noting it since it reads like coverage. The env-override tests are solid.

Happy to help land this — the code is basically ready; it's really just getting the OSS doc/example right so people don't end up with a bucket-less endpoint. Thanks again for the contribution!

synchrone:
You're completely right — thank you for the careful read. My original validation used `curl --aws-sigv4` against `https://<bucket>.oss-cn-hongkong.aliyuncs.com/...` (bucket in the host), and the doc example drifted to the bare region endpoint exactly as you guessed. `object_store` uses the endpoint verbatim when `virtual_hosted_style_request(true)`.

To be sure, I re-verified against a live scratch OSS bucket (`manual-nora-vhost-repro`, cn-hongkong) running this branch's binary as a Docker pull-through proxy (`NORA_DOCKER_PROXIES=https://mirror.gcr.io`, pulled `library/busybox:latest` through it):

| config | `/ready` | object writes |
|---|---|---|
| `vhost=false`, bare region endpoint | 503 | n/a — `403 SecondLevelDomainForbidden` (the bug this PR fixes) |
| `vhost=true`, bare region endpoint (old doc example) | **200 — false positive** | every PUT fails `403 PermanentRedirect` |
| `vhost=true`, `https://manual-nora-vhost-repro.oss-cn-hongkong.aliyuncs.com` | 200 | ✅ manifests + `.meta.json` sidecars persist, verified via signed `ListObjectsV2` |

Two sharp edges the misconfigured middle row exposes, now called out in the docs:
- `/ready` lies: the bucket-less `list(None)` hits the region root, which is OSS's **ListBuckets** API, and it parses — health reports fine while storage is broken.
- The PUT failure mode is nasty: OSS interprets the first key segment (`docker/`) as a *bucket name* — the error we got back was a `PermanentRedirect` for a foreign bucket named `docker` in cn-beijing.

Pushed in 36e39d1:
- OSS example + doc comments (`s3.rs`, `storage.rs`, settings EN/RU) now say the endpoint is used **as-is** and must include the bucket host, with the bucket-less-endpoint caution
- Replaced the constructor smoke test with two `wiremock` tests that pin the addressing contract: path-style lists `/test-bucket`, virtual-hosted lists `/` verbatim — these would have caught the doc drift, per your note

`cargo test --lib --bin nora` (1437 tests), `clippy --all-targets -- -D warnings`, and `fmt --check` are green.

ertime037 (approved):

devitway (approved):

devitway:
🤖 Relanded as a signed commit in #798 (thanks @synchrone — full authorship credited via `Co-authored-by`). `main` requires verified signatures and the branch commits were unsigned, so the reviewed diff was re-landed as one web-flow-signed commit. This PR will be closed once #798 merges.

devitway:
Landed on `main` via #798 (merged) — thank you, @synchrone! 🙏

Because `main` requires verified signatures and the branch commits were unsigned, the reviewed diff was re-landed as a single web-flow-signed commit. Your authorship is fully preserved via `Co-authored-by: Aleksandr Bogdanov` on the merged commit.

Closing this PR as relanded. Thanks again for the contribution and for iterating on the addressing-contract tests.

PR #809, opened 2026-07-04
docs(curation): document internal_namespaces pattern forms (#731)

### LINKED ISSUE #731: docs: internal_namespaces glob form — dotted/pathed names need `prefix.**` / `prefix/**`, not a single `*`
### Problem

`internal_namespaces` (the dependency-confusion defense, always active) glob-matches a package name against operator-supplied patterns via `glob_match` (`nora-registry/src/curation.rs`). The matcher special-cases the segment separators **only** under a `**` suffix:

- `prefix.**` matches `prefix`, `prefix.x`, **and** `prefix:x` (the Maven `group:artifact` colon)
- `prefix/**` matches `prefix`, `prefix/x/deeper` (the Terraform `ns/type`, Go module-path slash)
- a single trailing `*` is a plain `starts_with` and does **not** cross a `.` / `:` / `/` boundary in the intended way

So for registries whose internal name is **dotted or pathed** — Maven (`group.dotted:artifact`), Terraform (`ns/type`, `ns/name/provider`), Ansible (`ns.name`), Go (module path) — a pattern like `com.acme.*` or `acme/*` will **silently fail to match** `com.acme:lib` / `acme/aws`, and the dependency-confusion guard is bypassed without any error. The operator believes they are protected when they are not.

### Ask

Document the required glob form in the config reference (and on getnora.dev), next to `internal_namespaces` / `NORA_CURATION_INTERNAL_NS`:

- flat names (npm `@scope/*`, pypi/cargo/gems/nuget/conan/pub): `prefix*` or `@scope/*` are fine
- **dotted** names (Maven `group:artifact`, Ansible `ns.name`): use `group.**` / `ns.**`
- **pathed** names (Terraform `ns/type`, Go module paths): use `ns/**`

A short table mapping each registry to its name form + the recommended pattern would prevent the footgun.

### Context

Surfaced while closing the cross-registry metadata-isolation gap (getnora-io/nora-contrib-kit#68). The guards now cover every metadata path, but their effectiveness depends on the operator writing a pattern in the matching glob form. Anchor: `glob_match` in `nora-registry/src/curation.rs`.

Docs-only; no code change.

devitway:
Closes #731.

The **Internal Namespaces** section had `@mycompany/**` / `com.mycompany.**` examples but never documented the matcher, so `prefix.**` / `prefix/**` vs a single `*` was unclear. Adds a pattern-form table grounded in the actual matcher (`curation.rs` `glob_match`):

- `prefix/**` — namespace + all descendants, slash coordinates (docker, raw, npm scope, Go)
- `prefix.**` — namespace + all descendants, Maven dotted/colon (`group.artifact` / `group:artifact`)
- bare `name` — exact coordinate only (misses children)
- `prefix*` — raw prefix, ⚠️ over-matches siblings (`myorg*` also hits `myorganization`)
- `*` — everything

Plus a rule-of-thumb to prefer the boundary-aware `**` forms (dependency-confusion defense). Docs-only.

### DIFF
--- a/docs-site/src/content/docs/configuration/curation.md
+++ b/docs-site/src/content/docs/configuration/curation.md
@@ -195,6 +195,20 @@ internal_namespaces = ["@mycompany/**", "com.mycompany.**"]
 export NORA_CURATION_INTERNAL_NS="@mycompany/**,com.mycompany.**"
 ```
 
+### Pattern forms
+
+A pattern is matched against a package's **coordinate** — docker `org/image`, raw `org/path`, npm `@scope/name`, Go the module path, Maven the dotted/colon `group.artifact` / `group:artifact`, PyPI/Cargo the normalized name.
+
+| Pattern | Matches | Use for |
+|---------|---------|---------|
+| `prefix/**` | `prefix` itself **and** `prefix/anything/deeper` | slash-pathed coordinates: docker, raw, npm scope, Go |
+| `prefix.**` | `prefix`, `prefix.anything`, `prefix:anything` | dotted/colon coordinates: Maven `group.artifact` / `group:artifact` |
+| `name` (no wildcard) | that **exact** coordinate only — **not** its children | pinning a single package |
+| `prefix*` | any value **starting with** the literal `prefix` (crosses `/` and `.`) | ⚠️ over-matches — `myorg*` also matches `myorganization` |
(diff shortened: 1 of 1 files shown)


devitway:
Closing — docs-site/ is gitignored in the code repo; the getnora.dev docs are maintained in a separate repo, so this PR is empty. Will target the docs repo instead. (#731 stays open)

PR #834, opened 2026-07-11
ci: add diff-scoped coverage gate on pull requests

### LINKED ISSUE #598: CI: diff-scoped coverage gate on PRs (fail when new/changed lines are under-tested)
## Problem

Our coverage job doesn't actually gate anything mergeable:

- It runs **only on push to `main`** (`.github/workflows/ci.yml:105` —
  `if: github.ref == 'refs/heads/main' && github.event_name == 'push'`), i.e. **after** a PR is
  already merged. PRs are never coverage-checked.
- It computes a single **whole-repo** tarpaulin percentage and uses it to update a badge gist
  (`ci.yml:120-135`). The only threshold is `fail-under = 38` in `tarpaulin.toml`, far below our
  actual line coverage (~72%), so it effectively never trips.
- Nothing is **diff-scoped**. A PR can add a large block of untested logic and the whole-repo
  percentage barely moves, so CI stays green.

Net effect: new, untested logic merges clean. That's exactly how our worst regressions ship —
new code on the changed lines with no test exercising it (the class behind several of our recent
correctness fixes around fail-closed→fail-open behavior and proxy/integrity paths). A whole-repo
average can't catch that; a gate on the *changed lines* can.

## Proposed solution

Add a **PR-triggered, diff-scoped** coverage gate, alongside (not replacing) the existing badge job:

1. On `pull_request`, produce `lcov` from our existing coverage tooling
   (tarpaulin already emits it; `tarpaulin.toml` engine is `Llvm`).
2. Compute the set of added/changed `.rs` lines in the PR (`git diff` against the merge base).
3. Intersect lcov hits with those lines; **fail the job if coverage on new/changed lines < 70%**.
4. Skip cleanly for docs-only / non-Rust PRs (no changed `.rs` lines ⇒ pass).
5. Keep the whole-repo badge job exactly as-is (still push-to-main only).

Keep it a self-contained CI step (shell + a small lcov-vs-diff script) — no runtime dependency, no
external service, nothing compiled into the binary. Honors the "zero-dep" posture; this is pure CI.

## Acceptance criteria

- [ ] The gate runs on `pull_request` (and `merge_group`), not just push-to-main.
- [ ] A PR that adds an untested function (new-line coverage < 70%) **fails** the gate.
- [ ] A PR that adds the same function **with tests** passes.
- [ ] A docs-only / non-`.rs` PR passes (no changed Rust lines ⇒ skip, not fail).
- [ ] The threshold is configurable in one place; default 70% on new/changed lines.
- [ ] The existing whole-repo badge job is unchanged.
- [ ] Gate output names the under-covered files/lines so contributors can fix without guessing.

## Implementation notes / files

- `.github/workflows/ci.yml` — new `coverage-diff` job triggered on PRs; reuse the tarpaulin install +
  lcov output, add a diff-intersection step.
- `tarpaulin.toml` — ensure `lcov` is in `out` (currently `Json`, `Html`); raising the whole-repo
  `fail-under` is a separate, optional follow-up — this issue is specifically the **diff** gate.
- Small helper script (shell/python) to parse `lcov` + `git diff --unified=0` and compute new-line %.

## Test plan

- Unit-test the diff-intersection helper with a fixed lcov + a fixed diff (covered vs uncovered new lines).
- Dry-run the job on a throwaway PR that adds an untested fn → gate fails; add a test → gate passes.
- Confirm docs-only PR is skipped.

## Out of scope

- Raising the whole-repo `fail-under` floor (separate follow-up).
- Coverage dashboards / any standing service — this stays an ephemeral CI job.

---
_Branch:_ `ci/diff-coverage-gate`
_Commit:_ `ci: add diff-scoped coverage gate on pull requests`

01luyicheng:
Closes #598.

## What

Adds a PR-triggered, diff-scoped coverage gate alongside the existing whole-repo badge job (which is left untouched). The gate fails the build when coverage on new/changed `.rs` lines drops below a threshold (default 70%, configurable in one place: the `--threshold` default in `scripts/diff-coverage.py`).

## Why

Per #598: a whole-repo average barely moves when a PR adds untested logic, so CI stayed green even on regressions. Since #598 was filed, the `coverage` job was already moved to run on PRs and the `fail-under` floor was unmasked — this PR lands the remaining, diff-scoped core of the issue.

## Changes

- `tarpaulin.toml`: add `"Lcov"` to `out` so tarpaulin emits an lcov info file.
- `scripts/diff-coverage.py`: zero-dependency (Python 3 stdlib only) helper that parses the lcov file, computes the merge-base via `git merge-base`, enumerates new/changed `.rs` line ranges via `git diff --unified=0 --diff-filter=AM`, intersects them with the lcov `DA:` records, and fails when diff-scoped coverage is below the threshold. Docs-only / non-instrumentable PRs pass cleanly. Includes an in-script `--self-test` mode covering parsing, intersection, path normalization, and edge cases.
- `scripts/test-diff-coverage.sh`: self-test wrapper mirroring the existing `scripts/test-reland-filechanges.sh` convention.
- `.github/workflows/ci.yml`: new `coverage-diff` job (runs on `pull_request` and `merge_group`) that runs tarpaulin with `--fail-under 0` (so the diff-scoped gate is the sole arbiter of this job) and invokes the helper. A self-test step is also appended to the existing `coherence` job.

The existing `coverage` badge job is unchanged.

## Acceptance (#598)

- [x] Gate runs on `pull_request` **and** `merge_group` (the `if:` covers both; `BASE` is selected via a ternary: `github.event.pull_request.base.sha` on PR, `github.event.merge_group.base_sha` on merge_group)
- [x] PR adding an untested fn fails the gate (verified locally with a synthetic repo)
- [x] Same fn with a test passes
- [x] Docs-only / non-`.rs` PR passes (skip, not fail)
- [x] Threshold configurable in one place — the `--threshold` default in `scripts/diff-coverage.py`; the CI step does not pass `--threshold` explicitly
- [x] Existing whole-repo badge job unchanged
- [x] Gate output names under-covered files and line numbers

## Test plan

- `python3 scripts/diff-coverage.py --self-test` — all 16 cases pass (lcov parsing, diff-hunk parsing, intersection, path normalization, deletion handling, malformed input, line compaction)
- `bash scripts/test-diff-coverage.sh` — passes
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — YAML valid
- `actionlint` (v1.7.12, with the repo's own ignore flags from the `lint-workflows` job) — clean, exit 0
- Verified locally in a synthetic git repo: untested fn → fail (reports uncovered line); with test → pass; docs-only → skip/exit 0; missing lcov → exit 2; git error → exit 2 (no traceback)

The Rust gates (`cargo fmt --check`, `cargo clippy --package nora-registry -- -D warnings`, `cargo test --lib --bin nora`) are unaffected by this change; CI will validate them on the PR.

## Notes

- The `coverage-diff` job uses `--fail-under 0` so the global floor in `tarpaulin.toml` (60) does not preempt the diff-scoped gate. The existing `coverage` badge job still enforces the global floor.
- `fetch-depth: 0` is required for `git merge-base` / `git diff` to resolve the PR base ancestor (same pattern as the `security` and `semver` jobs).
- The helper uses `--src-prefix=a/ --dst-prefix=b/` on `git diff` so it is robust against a local `diff.noprefix = true` git config.
- DCO sign-off included in the commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,8 @@ jobs:
         run: ./scripts/verify-changelog.sh
       - name: Reland FileChanges builder — large-PR regression
         run: ./scripts/test-reland-filechanges.sh
+      - name: Diff-coverage helper self-test
+        run: ./scripts/test-diff-coverage.sh
 
   test:
     name: Test
@@ -149,6 +151,40 @@ jobs:
           valColorRange: ${{ env.COVERAGE }}
           minColorRange: 0
           maxColorRange: 100
+
+  coverage-diff:
+    name: Coverage (diff)
+    runs-on: ubuntu-latest
+    # Diff-scoped gate: fails the PR when the incremental coverage of changed
+    # .rs lines drops below the threshold, independent of the global fail-under
+    # floor enforced by the `coverage` job above (left untouched).
+    if: github.event_name == 'pull_request' || github.event_name == 'merge_group'
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          fetch-depth: 0
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # stable
+
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@42dc69e1aa15d09112580998cf2ef0119e2e91ae # v2
+
+      - name: Install tarpaulin
+        run: cargo install cargo-tarpaulin --locked
+
+      - name: Run coverage
+        run: cargo tarpaulin --config tarpaulin.toml --fail-under 0
+
+      - name: Diff-scoped coverage gate
+        env:
+          BASE: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.merge_group.base_sha }}
+        run: |
+          LCOV=$(find coverage -type f \( -name "*.lcov" -o -name "lcov.info" \) | head -1)
+          if [ -z "$LCOV" ]; then echo "No lcov file found" >&2; exit 1; fi
+          python3 scripts/diff-coverage.py --lcov "$LCOV" --base "${BASE}"
 
   security:
     name: Security
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Reland FileChanges builder — large-PR regression
         run: ./scripts/test-reland-filechanges.sh
       - name: Diff-coverage helper self-test
-        run: ./scripts/test-diff-coverage.sh
+        run: bash scripts/test-diff-coverage.sh
 
   test:
     name: Test
--- a/scripts/diff-coverage.py
+++ b/scripts/diff-coverage.py
@@ -1,0 +1,417 @@
+#!/usr/bin/env python3
+"""Diff-scoped coverage gate.
+
+Computes coverage over only the new/changed ``.rs`` lines in a pull request
+and fails the build when it drops below a threshold. Unlike tarpaulin's global
+``fail-under`` floor (which gates the whole-repo percentage), this gates the
+*incremental* coverage so a high-coverage repository does not regress
+line-by-line on each PR.
+
+Usage::
+
+    python3 diff-coverage.py --lcov coverage/lcov.info --base <merge-base-sha> \\
+        --threshold 70
+    python3 diff-coverage.py --self-test
+
+Inputs:
+    --lcov       Path to an lcov.info file (tarpaulin's ``"Lcov"`` output).
+    --base       The PR base SHA. The merge-base with HEAD is computed from it
+                 so rebased/multiple-commit PRs diff against the right ancestor.
+    --threshold  Minimum diff-scoped coverage percentage (default 70).
+
+The script invokes ``git`` (``git merge-base``, ``git diff``) to discover
+changed ``.rs`` line ranges, intersects them with the lcov ``DA:`` records, and
+reports per-file gaps. Docs-only PRs (no changed, instrumentable ``.rs`` lines)
+pass cleanly.
+
+Only changed lines that carry a ``DA:`` record are counted — non-instrumentable
+lines (comments, item declarations, ``use`` statements) are ignored, matching
+the behaviour of diff-cover style tools. A changed line with ``DA:<line>,0`` is
+uncovered; ``DA:<line>,<n>`` with ``n > 0`` is covered.
+"""
+
+import argparse
+import os
+import re
+import subprocess
+import sys
+
+# Matches a unified-diff hunk header: @@ -<old>,<old_n> +<new>,<new_n> @@
+# --unified=0 emits only added lines, but we read the new-side range from the
+# header so a brand-new file (@@ -0,0 +1,N @@) is handled the same way.
+_HUNK_RE = re.compile(r"^@@ -\d+(?:,\d+)? \+(\d+)(?:,(\d+))? @@")
+
+
+def parse_lcov(text):
+    """Parse lcov text into ``{filepath: {line: hit_count}}``.
+
+    Only ``SF:`` and ``DA:`` records are used. ``DA:<line>,<count>[,<checksum>]``
+    gives the hit count for a line; count > 0 means covered, 0 means uncovered.
+    """
+    files = {}
+    current = None
+    for line in text.splitlines():
+        if line.startswith("SF:"):
+            current = line[3:].strip()
+            files.setdefault(current, {})
+        elif line.startswith("DA:") and current is not None:
+            parts = line[3:].split(",")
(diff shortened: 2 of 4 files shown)


devitway (commented):
Thanks for this, @01luyicheng — solid, well-scoped work. I ran the helper's `--self-test` locally (16/16 pass) and read the full path; the engineering is careful. A few notes, none blocking.

## What's good
- **Safe trigger:** the job runs under `pull_request`/`merge_group` (not `pull_request_target`), and further restricts `permissions: contents: read`.
- **Fail-closed:** missing lcov → `exit 1`, and the script's exit codes (1 = under threshold, 2 = error) propagate with no `|| true` / `continue-on-error` — the gate can't be silently skipped.
- `fetch-depth: 0`, so `git merge-base <base> HEAD` resolves on rebased / multi-commit PRs.
- Actions pinned by SHA; the self-test is wired non-vacuously (`set -euo pipefail`) and covers good edge cases (docs-only PR, non-instrumentable lines, `/dev/null` deletions, malformed `DA:`, absolute-path normalization).

## Notes

1. **Excluded / non-instrumented lines are ungated — worth a doc line.** `tarpaulin.toml` excludes `ui/*`, `main.rs`, `openapi.rs`, so changed lines there emit no `DA:` records and `intersect()` counts them as zero instrumentable — i.e. a PR adding untested logic in those paths passes the diff gate. Same for `#[cfg(...)]` code not built in the coverage profile. This is inherent to diff-coverage tooling, and you already note "a changed file with no lcov data contributes zero," but a one-liner in the PR/README calling out that the excluded paths are ungated-by-design would set expectations.

2. **The job runs but won't block merge until it's a required check.** Adding `Coverage (diff)` here makes it execute, but branch protection must list it as required for it to actually gate (can't be done from the PR). Flagging so it isn't assumed enforcing on day one — the `merge_group` trigger suggests that's the intent.

3. **Tarpaulin now runs twice per PR.** The existing `coverage` job and this new `coverage-diff` job each run `cargo tarpaulin` (full instrumented build). Sharing the lcov via an artifact / `needs:` would roughly halve the coverage cost per PR. Optional.

Nits: `_lookup_hits`'s suffix fallback could in theory match the wrong file if two paths shared a suffix (exact match is tried first and repo paths are unique, so negligible); and the 70% threshold lives in the script default rather than being passed in `ci.yml`, so tuning it means editing the script.

Nice addition overall — thanks again.

ertime037 (dismissed):

devitway:
🤖 Relanded as a signed commit: https://github.com/getnora-io/nora/pull/835 (thanks @01luyicheng — full authorship credited via `Co-authored-by`). This PR will be closed once the reland merges.

devitway:
Heads-up on the failing `Coherence` check — the `Diff-coverage helper self-test` step fails with `./scripts/test-diff-coverage.sh: Permission denied` (exit 126), but **not because of anything in your PR**: your `scripts/test-diff-coverage.sh` is correctly `100755` (executable).

`main` here requires verified commit signatures, so this PR lands via an auto-reland that re-commits the reviewed diff through GitHub's `createCommitOnBranch` API. That API's `FileChanges` carries only `{path, contents}` — no file mode — so **newly added files land as `100644`, dropping the executable bit**. The relanded `test-diff-coverage.sh` ends up non-executable and `./scripts/test-diff-coverage.sh` then fails with "Permission denied" on the runner. (Existing scripts like `test-reland-filechanges.sh` keep their bit because the reland doesn't recreate them.)

Simplest fix that's robust to this — invoke the self-test mode-independently in `ci.yml`:

```yaml
- name: Diff-coverage helper self-test
  run: bash scripts/test-diff-coverage.sh
```

`bash scripts/…` runs regardless of the executable bit and survives the reland (worth doing for the reland self-test line too, for consistency).

Thanks again — the logic itself is solid; this is purely a reland-mechanism quirk on our side, which we're tracking separately.

ertime037 (approved):

devitway:
🤖 Relanded as a signed commit: https://github.com/getnora-io/nora/pull/836 (thanks @01luyicheng — full authorship credited via `Co-authored-by`). This PR will be closed once the reland merges.

devitway:
Thanks a lot, @01luyicheng — the diff-scoped coverage gate is now on `main` via #836, relanded as a single signed commit (`main` requires verified signatures; your review and full authorship are preserved via `Co-authored-by`). The self-test invocation was switched to a mode-independent `bash scripts/…` call so it survives the reland's executable-bit handling. Closing in favor of the merged reland. 🎉

devitway:
Credit where it's due, @01luyicheng — the diff-scoped coverage gate is entirely your design and implementation (#834, shipped via #836).

Unfortunately the git `Co-authored-by` attribution on the merged commit got dropped: our auto-reland credits only the **head** commit's author, and after I pushed a one-line CI fix on top of your branch, the head became my commit (and a self-co-author is de-duplicated by GitHub) — so your authorship didn't carry through to `main`. The merged history can't be rewritten, but the work is yours and I want that on record.

We're fixing the reland to credit **every** commit author in a PR so this can't recur. Thanks for the contribution — it's a genuinely useful gate.

PR #835, opened 2026-07-11
ci: add diff-scoped coverage gate on pull requests

devitway:
Signed reland of #834 by @01luyicheng.

\`main\` requires verified signatures; the original commits were unsigned, so this re-lands the reviewed diff as one web-flow-signed commit.

Original PR: #834
Signed commit: https://github.com/getnora-io/nora/commit/8451c976da5777f267702429cd6e01ea84c794f2


devitway:
Superseded by #836 — re-relanded after the diff-coverage self-test was switched to a mode-independent `bash scripts/…` invocation (the reland's createCommitOnBranch drops the executable bit, so `./scripts/…` failed Coherence here). Closing.

Issue #128, opened 2026-04-12
Add deb/rpm package repository


addspin:
### Problem

Hello!
Thanks for your work, I love all-in-one solutions!
I would like to emphasize the importance of data storage registries as a top priority. Could you also include the deb/rpm package repository in the list? This is crucial for ensuring the stability and security of all closed segments.

### Proposed Solution

Include the deb/rpm package repository

### Related Registry

Core/General

makavity:
@devitway i can implement it by `cargo-deb` and `cargo-generate-rpm` if it's fine for you.

devitway:
@makavity this issue is about NORA acting as an APT/YUM **repository server** (hosting .deb/.rpm packages for consumers). That's a post-1.0 feature.

Packaging the NORA binary itself as .deb/.rpm is a different thing — and a welcome one. Created #209 for that and assigned you. `cargo-deb` + `cargo-generate-rpm` is exactly the right approach.

devitway:
Update — the **hosting** side of this has shipped and is on `main` (part of the upcoming 1.0):

- **APT** at `/deb/` — each `/deb/{repo}/` is an independent flat repo; `PUT`/`DELETE` regenerate `Packages`/`Packages.gz`/`Release`. Verified end-to-end against `apt`.
- **YUM/DNF** at `/rpm/` — each `/rpm/{repo}/` regenerates `repodata/` (repomd + primary/filelists/other). Verified end-to-end against `dnf` on Fedora 41.

Both are hosted-only, default-disabled, control/header parsed server-side (pure-Rust, decompression-bounded), rebuilt under the per-repo publish lock, fail-closed.

**Remaining (genuinely post-1.0): GPG-signed indexes** — `InRelease` / `Release.gpg` for APT and signed `repomd.xml` for YUM. Today indexes are unsigned and clients use `[trusted=yes]` / `gpgcheck=0 repo_gpgcheck=0`. Retitling this issue to track that.

jgrund:
> Remaining (genuinely post-1.0): GPG-signed indexes — InRelease / Release.gpg for APT and signed repomd.xml for YUM. Today indexes are unsigned and clients use [trusted=yes] / gpgcheck=0 repo_gpgcheck=0. Retitling this issue to track that.

Working on this now.

devitway:
[...]

PR #841, opened 2026-07-12
feat: GPG-signed rpm/deb repository indexes (#128)

jgrund:
## What

Adds OpenPGP signing of rpm/deb repository indexes — the remaining half
of #128 after hosting shipped:

- New `signing.rs`: one per-instance signing key (v4 Ed25519, the
  variant every deployed gpgv/gnupg verifier understands — the RFC 9580
  v6 algorithms are not yet recognized by either), generated at first
  boot and persisted armored with owner-only permissions via
  write-tmp-then-rename. A present-but-corrupt key is a fatal startup
  error: silently rotating would invalidate every client's pinned key.
- rpm: every repodata regeneration writes a detached armored
  `repodata/repomd.xml.asc` (after repomd.xml, so a reader never sees a
  signature for bytes that are not there yet); the armored public key
  is served at `repodata/repomd.xml.key` for `gpgkey=`. Clients verify
  with `repo_gpgcheck=1`.
- deb: every index regeneration writes a clearsigned `InRelease` and a
  detached `Release.gpg`; the public key is served at `pubkey.gpg` for
  an apt `signed-by` keyring — no more `[trusted=yes]`.
- Fail-closed: signing errors abort the regeneration like any other
  rebuild error. Turning signing off removes stale signatures on the
  next regeneration (a leftover mismatched signature would hard-fail
  apt/dnf). NORA signs metadata, not packages — `gpgcheck=0` remains
  for the packages themselves.
- Config: `[signing] enabled` (default true) + `key_path`. Local
  storage derives `<storage.path>/.signing/nora.key`; S3 requires an
  explicit path (warned + disabled otherwise, since there is no local
  data directory to keep a generated key in). Env:
  `NORA_SIGNING_ENABLED`, `NORA_SIGNING_KEY_PATH`.
- New dep: `pgp` (rpgp) 0.20, MIT OR Apache-2.0, pure Rust (no
  openssl), default-features off; `rand` 0.8 to match its rng
  requirement. cargo-deny green.

Also repairs a broken code fence in COMPAT.md's rpm `.repo` example
that slipped through the deb-registry merge.

## Why

Signed indexes were the tracked remainder of #128: without them every
client had to opt out of verification (`[trusted=yes]`,
`gpgcheck=0 repo_gpgcheck=0`), which is unacceptable for the
closed/air-gapped segments the feature targets.

## Checklist

- [x] Tests pass (`cargo test`) — 1517 pass, incl. cryptographic
      verification of every signature shape against the served public
      key. End-to-end with verification enforced: dnf `repo_gpgcheck=1`
      + `gpgkey=` (Fedora 41) and apt `signed-by` without
      `[trusted=yes]` (Debian 12) both install; both hard-reject a
      wrong key (dnf: "GPG signature verification error", apt:
      NO_PUBKEY).
- [x] New registry? See CONTRIBUTING.md checklist (not a new registry;
      OpenAPI stubs added for both public-key endpoints)


devitway (commented):
Thanks for this — it's a well-built completion of the #128 signing story, and the crypto fundamentals are right: v4 Ed25519 (the variant deployed `gpgv`/`librepo` actually verify), fail-closed key load that never silently rotates a pinned key, atomic `0600` persist, signing done inside the per-repo `publish_lock` over the exact bytes that get served, default-on, and no daemon/DB. Nice tests around reload-keeps-identity and corrupt-key-fails-closed.

I'd like two things addressed before merge, plus a few optional hardenings.

## Please address before merge

**1. Multi-replica key divergence — the docstring claims it works, but it doesn't for the HA/S3 case.**
`signing.rs:14-16` says *"a restart — or a replica sharing the path — keeps the same identity."* That holds for a single writer on a shared volume, but the signing key is always on the **local** filesystem: for S3 storage `key_path` is required and the key is never stored in S3 (`main.rs:936-946`, `config/signing_cfg.rs:16-18`). S3 is exactly the multi-replica story — so with N pods on per-pod/ephemeral storage, each pod generates a **different** key on first boot. A client that imports `pubkey.gpg` from pod A and then fetches an `InRelease` signed by pod B (round-robin LB) gets a hard verification failure.

Could you document this as a known limitation rather than stating it works — i.e. the signing key is single-writer, and multi-replica requires the same key bytes on every replica (a shared `ReadWriteMany` volume for `key_path`, or a key provisioned once and mounted read-only into every pod, e.g. a K8s Secret) — with an upgrade path noted? A `WARN` at startup when `storage == S3` and the key resolved to a node-local path would also help operators catch a mis-provisioned setup.

**2. The tests prove rPGP round-trips, not that `gpgv`/`dnf` accept the output.**
Every `verify()` in the suite calls rPGP's own verifier against the rPGP-generated key (`signing.rs:174-194`). That proves internal consistency but not client acceptance — and the gaps that only real verifiers catch are exactly the ones that bite (apt's `Date:`/field-layout and weak-digest policy, InRelease dash-escaping/trailing-newline canonicalization, librepo's armor/key-flag expectations). Could you add one integration test (feel free to gate it behind a feature or `#[ignore]` if the CI image lacks the tools) that runs the real verifiers over the **served** bytes: `gpgv` on `Release`/`Release.gpg` and on `InRelease`, and `gpg --import repomd.xml.key && gpg --verify repomd.xml.asc repomd.xml`? That's the only thing that closes the round-trip-vs-acceptance gap.

## Optional (non-blocking) hardening

- **First-boot generation race** (`signing.rs:42-56`): `read → on NotFound generate + persist` isn't atomic across concurrent starts. Two processes that both see `NotFound` each generate a key; the atomic rename keeps the file uncorrupt, but the process that lost the rename keeps a *different* in-memory key and serves a mismatched public key for its lifetime. Consider create-exclusive (`O_EXCL`) on the final path and, on `AlreadyExists`, discard the fresh key and re-read the winner.
- **Zeroize the secret** (`signing.rs:44,116`): the armored private key is read/built as a plain `String`; the codebase already uses `zeroize`. Wrapping both intermediates in `Zeroizing<String>` keeps the secret from lingering in freed heap.
- **S3 + enabled + empty `key_path` serves unsigned with only a `warn`** (`main.rs:938-944`), while the default is `enabled=true` and COMPAT/README advertise signing as "Full" (`COMPAT.md:262`). Consider either a hard boot error (force the operator to set `key_path` or explicitly disable) or a COMPAT footnote that S3 requires `NORA_SIGNING_KEY_PATH`.

## Nits

- No key expiry (`signing.rs:101-110`) — the conventional choice for a pinned repo key; you might ship a revocation certificate at generation time so a leaked key can be revoked without server access.
- `deny.toml` RUSTSEC-2023-0071 (rsa Marvin) ignore comment now also applies to `pgp` (which links `rsa`); the ignore stays correct since signing is Ed25519 (no rsa private-key op), worth a one-line note.
- `repomd.xml` is written before its `.asc` (`rpm.rs:636` then `:650`) — a reader between the two puts sees the new index with a stale/absent signature; same transient class as the existing "old blob" window and resolved by a client retry, but worth acknowledging in the comment.

Happy to re-review once the two above are in. Thanks again — this is a solid contribution.

jgrund:
Both must-fix items plus the hardenings addressed in 4256dcd:

**1. Multi-replica honesty** — module docs now state the key is single-writer and spell out the multi-replica requirement (same key bytes on every replica: read-only Secret mount or RWX volume). Startup WARNs when the key was *generated* on this boot while storage is an object store — that's the detectable signature of a mis-provisioned ephemeral key path (couldn't reliably detect "node-local" itself, but generation-on-S3/GCS is the case that bites).

**2. Real-verifier acceptance test** — `real_verifier_accepts_all_signature_shapes` (`#[ignore]`, needs gpg+gpgv on PATH): imports the served pubkey, `gpgv` over Release/Release.gpg and over InRelease, `gpg --verify` over repomd.xml.asc. Green against GnuPG 2.4 locally; same shapes were also accepted end-to-end by dnf (`repo_gpgcheck=1`, Fedora 41) and apt (`signed-by`, Debian 12) in container runs during development, including wrong-key hard rejections.

**Hardenings taken:** create-exclusive first-boot claim (losers adopt the winner's key) — and the 8-thread convergence test you prompted caught a real second-order bug: my staging temp name was per-process, so same-pid concurrent callers could cross-publish each other's staged bytes; staging is now unique per call. `Zeroizing` on all armored-secret intermediates. COMPAT footnote for the S3/GCS `key_path` requirement — kept warn-not-error so existing S3 rpm/deb deployments don't break on upgrade, with the WARN making the unsigned state visible. deny.toml note extended to `pgp`. rpm.rs comment acknowledges the repomd→.asc reader window.

**Deferred with rationale:** key expiry + revocation certificate — agreed on the direction; deferring so expiry/rotation/re-signing land as one design (rotation needs the re-sign-all-repos sweep that's queued behind the reindex endpoint).

devitway (commented):
Thanks for the thorough write-up — taking a look now. The create-exclusive first-boot claim + per-call staging fix and the gpgv/dnf/apt acceptance coverage all read well; the multi-replica WARN-on-generate is a sensible detectable signal. Fine deferring key expiry/revocation as one rotation design. Will follow up with anything after I finish reading through.

ertime037 (approved):

PR #842, opened 2026-07-12
feat: native Google Cloud Storage backend (storage.mode = gcs)

jgrund:
## What

Generalizes the S3 backend into `ObjectStorage` over the `object_store`
trait (S3 path unchanged, file renamed s3.rs → object.rs) and adds a
GCS constructor:

- `storage.mode = "gcs"` / `NORA_STORAGE_MODE=gcs`; bucket from the
  existing `storage.bucket`.
- Credentials resolve as: explicit service-account JSON
  (`storage.gcs_service_account_path` /
  `NORA_STORAGE_GCS_SERVICE_ACCOUNT_PATH`), then ambient `GOOGLE_*`
  env, then the instance metadata server — GKE Workload Identity and
  GCE service accounts work with no key material.
- `storage.gcs_base_url` overrides the endpoint for emulators or
  Private Google Access; an `http://` override also skips request
  signing (emulators don't verify signatures).
- `nora migrate` accepts `gcs` as source/destination; /health reports
  `backend: gcs`; bucket-must-not-be-empty validation covers gcs.
- Same single-writer caveat as S3 for rpm/deb publishing (in-process
  publish lock) — documented in COMPAT.md.

## Why

GCP deployments previously had to use GCS's S3-interoperability layer
with static HMAC keys — no Workload Identity, and S3-conditional-write
semantics don't translate cleanly through interop. Native mode removes
the static credentials and uses GCS's own API.

## Checklist

- [x] Tests pass (`cargo test`) — 1511 pass. Verified end-to-end
      against real GCS (project bucket, ADC): /health reachable,
      raw roundtrip, rpm publish + dnf install, deb publish + apt
      install, DELETE regeneration. (fake-gcs-server can't serve
      object_store's XML-API PUT — fsouza/fake-gcs-server#994 — so
      verification used real GCS.)
- [x] New registry? See CONTRIBUTING.md checklist (not a registry)

Note: overlaps #841 only in main.rs/test_helpers.rs initializer lines — whichever merges second has a trivial conflict.


devitway (commented):
Thanks for this — it's a clean, well-scoped generalization: the abstraction is genuinely over the `object_store` trait, so renaming `s3.rs` → `object.rs` and adding a parallel `new_gcs` is the right shape rather than a bolt-on. `new_gcs` faithfully mirrors `new_s3` (lazy builder, ADC ladder via `from_env`, optional service-account path, emulator `base_url`), the struct is renamed consistently to `ObjectStorage`, `pin_store: None` matches S3's at-rest posture, and COMPAT already carries the GCS row. The design is good.

One real thing to address before merge (tests), plus the CI red is partly a tooling artifact on our side — details below so you don't chase the wrong lines.

## The red `Coverage (diff)` check — two separate things

**1. Most of the reported miss is a diff-coverage bug on our side, not your code.**
`scripts/diff-coverage.py` runs `git diff --no-renames`, so your `s3.rs` → `object.rs` rename (git similarity `R088`, i.e. 88% identical) is scored as a brand-new 567-line file. That drags in the pre-existing async `StorageBackend` I/O methods (`put`/`get`/`delete`/`stat`/`get_reader`) — which have always been integration-only and were never counted on `main` — and reports them as newly-uncovered. That's where most of the `41/79 (51.9%)` came from. A fix to make the gate rename-aware (`--find-renames`) is being prepared separately; you don't need to do anything about those grandfathered lines.

**2. After that, there is still a real, small gap: the new GCS paths are untested.**
With the rename correctly scored, the genuinely-new uncovered lines are just `new_gcs` (`object.rs:83-107`) and the `Storage::new_gcs` dispatch (`storage/mod.rs`) — 0/14 and the wrapper. This one is worth fixing in the PR, and it's easy: the existing `new_s3` path is already **17/17 covered** by `test_s3_storage_creation_anonymous`, which constructs `ObjectStorage::new(...)` against a fake `http://localhost:9000` — no live S3. `new_gcs` builds identically without credentials or network: with no service-account path and no ADC file, `GoogleCloudStorageBuilder::build()` resolves credentials to a lazy `InstanceCredentialProvider` (no build-time I/O), so a sibling construction test works in CI with no emulator and no creds. Something like:

```rust
#[test]
fn gcs_backend_constructs_against_emulator() {
    let s = ObjectStorage::new_gcs("test-bucket", None, Some("http://localhost:4443"));
    assert_eq!(s.backend_name(), "gcs");
}
```

plus a `Storage::new_gcs(...)` wrapper test, covers `83-107` + the dispatch and takes the diff-scoped gate green honestly.

## Nits (non-blocking)

- `encode_s3_key` / `encode_s3_key_legacy` are now misnomers — they apply to every `object_store` backend, GCS included. A cosmetic `encode_object_key` rename would read truer.
- Worth one explicit COMPAT/README line that the `pin_store: None` at-rest-integrity posture (hash-pinning unavailable) applies to object stores generally, not only S3 — right now it's implied by "same as S3".
- `builder.build().expect("Failed to build GCS client")` panics on bad config. It matches the existing S3 `.expect` fail-fast, so it's consistent — but a `Result`-returning constructor would be the nicer long-term shape for both.
- The `let allow_http = url.starts_with("http://")` heuristic silently also sets `skip_signature(true)`, so any `http://` `base_url` disables request signing. This mirrors the existing `new_s3` behavior and only triggers when `base_url` is explicitly set (production `https://storage.googleapis.com` is unaffected), so it's fine — worth a one-line doc/config comment that `base_url` is emulator-only, since an operator pointing it at a plaintext endpoint would get unsigned requests.
- `migrate` reads `config.storage.bucket` for both the `s3` and `gcs` branches (`main.rs:801/822`), so an S3→GCS migration is same-bucket-name only — consistent with the pre-existing single-`bucket` assumption, but there's no way to migrate into a differently-named GCS bucket. Minor; flag it in the migrate docs if that's a real use case.

## What I checked (all clean)

To de-risk the rename for you: the `StorageBackend` impl is behavior-preserving — the only changed line in all 12 method bodies is `backend_name()` now returning `self.name` (which is `"s3"` for the S3 constructor), and `encode_s3_key`/`decode_s3_key`/`map_err` are byte-identical to `main` (so GCS inherits the exact S3 key-encoding and error-mapping, and the existing `s3_key_roundtrip` proptest still covers it). Config is fail-closed on an unknown `storage.mode` (exhaustive match, no wildcard default). Credentials look right: only the service-account *path* passes through — key material never touches config/env/logs.

Happy to re-review once the GCS construction tests are in. Nice contribution.

jgrund:
Review addressed in 6b68abc:

- **GCS construction tests**: `new_gcs` against an emulator `base_url`, against the default endpoint (both credential-lazy, no network), and the `Storage::new_gcs` wrapper asserting backend name + pin store disabled — covers `object.rs` GCS lines and the dispatch.
- **`encode_object_key` rename** taken, with a doc note that the scheme is shared across backends.
- **COMPAT**: pin-store posture now stated as applying to all object-store backends.
- **`gcs_base_url` doc**: explicit emulator-only warning that `http://` also disables request signing.

Not taken (agree with your framing they're fine as-is): `.expect` construction (kept consistent with S3; happy to do the `Result` refactor for both in a follow-up), migrate same-bucket-name constraint (flagging in migrate docs if a real cross-name use case appears). Thanks for chasing down the rename-blind diff-coverage artifact — good to know most of the red wasn't real.

ertime037 (dismissed):

devitway:
Heads-up — #841 (GPG signing) just merged, so `main` moved and this PR now shows a conflict. Only one file actually conflicts: `CHANGELOG.md`, where #841's GPG bullet and this PR's GCS bullet land at the same `## [Unreleased] → ### Added` spot. Everything else auto-merges cleanly — including the `s3.rs` → `object.rs` rename (`main` never touched `s3.rs`, so it's a clean rename-delete, not a conflict).

A rebase on `main` resolves it — keep both bullets.

One separate thing to fix while you're in there: the changelog entry for the GCS backend is currently **duplicated** on this branch — two identical `### Added` blocks with the same bullet, one ending `(#128 follow-on)` and one `(#844)`. Please collapse to a single bullet. So the top should end up as:

```
### Added
- **GPG-signed rpm/deb repository indexes** … (#128).
- **Native Google Cloud Storage backend (`storage.mode = "gcs"`)** … (#844).
```

Once that's pushed the conflict clears and it's back to green. Thanks!

ertime037 (approved):

devitway (approved):
GCS backend LGTM — construction tests, encode_object_key rename, and the pin-store/at-rest COMPAT note all addressed. The S3 path is behavior-preserving. Approving for 1.0.1.

PR #845, opened 2026-07-13
feat: streaming raw uploads and downloads — bounded memory at any size

jgrund:
## What

Raw uploads and downloads no longer buffer object bodies in memory.

Uploads (`PUT /raw/{path}`):
- The request body streams frame-by-frame to a temp file on the storage
  filesystem (O(frame) peak RAM); `raw.max_file_size` is enforced
  incrementally as frames arrive, plus a fast reject on an oversized
  declared Content-Length. `server.body_limit_mb` no longer applies to
  raw uploads.
- The streamed file is hashed once and committed with `put_from_path` —
  a same-filesystem rename on the local backend, streaming multipart on
  object stores — with the integrity pin recorded from the streamed
  hash, so the ETag/conditional-PUT flows are unchanged.
- Aborted/oversized/crashed uploads leave no temp file (RAII guard,
  covered by tests).

Downloads (`GET /raw/{path}`):
- Served via the storage reader with STREAMING integrity verification:
  frames are hashed as they are sent and, when a pin exists, the final
  frame is withheld until the digest is checked at EOF — a mismatch
  aborts the body one frame short of completion, so a tampered object
  can never arrive at the client as a complete download (fail-closed,
  observable as a Content-Length shortfall / transfer error). Unpinned
  keys are served without a cryptographic check, matching the buffered
  gate's Unpinned arm.
- Design note for review: this replaces raw's use of the buffered
  `get_verified` typestate gate with a streaming equivalent
  (`VerifyingStream`). The gate's guarantee is preserved in spirit —
  no complete corrupt download — but the compile-time witness is not
  carried through the stream; flagging explicitly since verified.rs is
  a deliberate design pilot.

Shared plumbing: the Docker blob-upload streaming helpers
(`stream_body_to_file`, `StreamOutcome`, `TempFileGuard`,
`content_length`) moved to shared registry code, plus a shared
`sha256_of_file` that Docker's digest verification now reuses.

Verified locally end-to-end: a 3 GiB object round-trips byte-identical
(upload 201, download 200, equal sha256) with the server's peak RSS at
37 MB across both transfers; a tampered pinned object is aborted
mid-body (test-covered).

## Why

`raw.max_file_size` was previously capped in practice by RAM (body
fully buffered) and by `server.body_limit_mb` (default 2 GiB), and
downloads of objects that large failed outright. Raw storage should
handle artifacts of any size — VM images, large tarballs — bounded by
disk, not memory.

## Checklist

- [x] Tests pass (`cargo test`) — 1601, incl. streamed roundtrip,
      mid-stream size-cap rejection with temp hygiene, Content-Length
      fast reject, tamper-abort
- [x] New registry? See CONTRIBUTING.md checklist (not a new registry)


devitway (commented):
Really solid PR — I read the full streaming path and the claims hold up.

**Verified (by code + your tests):**
- **Bounded memory** — `stream_body_to_file` keeps one frame in RAM and checks the budget *before* the write, so an over-budget frame is never written to disk. Covered by the 413-midstream test.
- **Temp hygiene** — `TempFileGuard` reclaims the temp file on every exit path (413, mismatch, client-gone, panic, early return). Covered (0 leftovers).
- **Streaming download integrity** — withholding the final frame until the EOF digest check is a clean way to keep streaming while staying fail-closed: a tampered object can never arrive complete. Covered by the tamper-abort test.
- **Same-FS temp** (`<storage>/tmp/…`) → `put_from_path` commits by rename, no EXDEV in the normal path.
- **docker.rs** is a behavior-preserving extraction of the shared helpers into `registry/mod.rs`, plus the `validate_upload_uuid` and `ProxyDownloadGuard` hardenings. `publish_lock` is taken *after* streaming — nice, a slow client no longer holds the key's lock.

**One thing worth adding — a durability fsync (and it's our gap, not yours):**

`stream_body_to_file` does `file.flush().await` but never `sync_all()`. `put_from_path`'s local direct-rename branch explicitly *"relies on the caller having fsync'd src"* — and nothing in the raw/docker upload path does. So after a power loss right after commit, a key can exist (its dir entry is synced) with data that never reached disk.

This isn't introduced here — it's been latent in docker blob uploads since #819, when switching to streaming for bounded memory silently dropped the `sync_all()` that the buffered `put()` path still does. Your PR extracts that same helper for raw, so it's the natural place to fix both: one `file.sync_all().await?` after the `flush()` in `stream_body_to_file` closes it for docker *and* raw.

Tracking the latent docker side in #846.

**Nit (non-blocking):** the `tmp/{raw,docker}-uploads` dirs aren't excluded from `total_size()`, so an in-flight upload transiently inflates the `storage_bytes` gauge — cosmetic.

jgrund:
Both taken in the follow-up commit:

- **fsync**: `stream_body_to_file` now `sync_all()`s after the final flush, with a comment tying it to `put_from_path`'s rename-relies-on-synced-source contract — closes it for raw and the latent docker case in one place (#846).
- **Gauge nit**: took it too — local `total_size()` now skips the top-level `tmp/` staging dir, so in-flight uploads don't sawtooth the storage gauge.

Thanks for the careful read — especially for tracing the docker durability gap to its origin.

ertime037 (approved):

devitway (approved):
Both follow-up fixes verified in the code: stream_body_to_file now sync_all()s after flush (closes the durability gap for raw and the latent docker case, #846), and total_size() skips the top-level tmp/ staging dir. Bounded-memory streaming + temp hygiene + streaming download integrity all hold. Thanks — nice work.

PR #847, opened 2026-07-13
feat: rpm/deb repository reconcile endpoint (POST /{repo}/-/reindex)

jgrund:
## What

Storage is the source of truth (ADR-2), but rpm/deb indexes only
regenerated inside the publish/delete handlers — a repository whose
storage changed out-of-band (direct deletion, bulk restore, objects
dropped straight into the bucket) served stale signed indexes with no
way to rebuild short of a dummy publish. This adds the reconcile verb:

- `POST /rpm/{repo}/-/reindex` and `POST /deb/{repo}/-/reindex`, under
  the repo publish lock, namespace-scope enforced, following the raw
  registry's existing `-/reindex` convention:
  - sidecars whose package is gone are dropped (heals out-of-band
    deletes — indexes stop advertising ghosts);
  - packages with no sidecar are parsed and adopted (heals out-of-band
    additions), with the same validation as the upload path — an
    invalid file fails the reconcile with 422 instead of being
    silently skipped into a repo that lies about its contents;
  - indexes are rebuilt and re-signed; JSON body reports counts.
- 404 for a repo with nothing in storage (a typo'd name must not
  create an empty repo).
- Doubles as the re-sign hook after a signing-key change: reindex each
  repo and clients verify against the new key.
- Groundwork: `regenerate_repodata` / `regenerate_indexes` no longer
  take request state — they run from storage + signer alone, so future
  background callers (retention over rpm/deb) can rebuild without a
  request in flight.

## Why

Operators interact with storage directly — the filesystem-is-the-
database design invites it — and both hosted formats had no recovery
path from that. This closes the gap and is the base for retention.

## Checklist

- [x] Tests pass (`cargo test`) — 1609, incl. heal-delete, adopt-add,
      invalid-package 422, unknown-repo 404, scope enforcement, for
      both formats. Verified live: a package removed from disk behind
      the API disappears from repodata after reindex and dnf accepts
      the re-signed repo with repo_gpgcheck=1.
- [x] New registry? See CONTRIBUTING.md checklist (not a new registry;
      OpenAPI stubs added for both endpoints)


ertime037 (approved):

devitway (approved):
Reviewed the reconcile path (rpm + deb): runs under the repo publish lock, namespace-scope enforced, repo name validated. Regeneration is correctly ordered — data files, then the index (repomd/Release) last, then the signature — so a reader never sees an index referencing unwritten files (same transient window as publish, not a new gap). Re-signs after regen, fail-closed on an unparseable package, idempotent. Nice foundation for the eventual key-rotation re-sign sweep. LGTM.

PR #848, opened 2026-07-13
feat: retention over rpm, deb, and raw with per-repo rule targeting

jgrund:
## What

Retention previously skipped rpm, deb, and raw silently — no collector
existed, so rules over those formats did nothing. This adds:

- **rpm/deb collectors** driven by the metadata sidecars (package
  payloads are never read). Versions group per `{repo}/{package}`; a
  version's keys are the package file and its sidecar, so deletion
  removes both. After deletions, every touched repo's indexes are
  rebuilt and RE-SIGNED under the same per-repo publish lock the
  handlers use — retention can never leave a signed index advertising
  deleted packages. A failed rebuild logs loudly and points at
  `-/reindex` (the deletions themselves are already durable).
- **raw collector** grouping depth-2 path prefixes
  (`raw/{name}/{version}/…`) as the aging unit — a directory of
  related files (image + checksums) ages out as one version. Files at
  the raw root are never collected, so untargeted uploads can't be
  swept accidentally.
- **`name_glob` on retention rules** — optional glob against the group
  name within a registry (`myrepo/*`, `*-dev-*/*`). First matching
  rule wins; specific rules go first; a group matching no rule is
  still kept forever. Enables shape-based policies: age-only for dev
  repositories, keep-last windows for rolling streams, no rule for
  release repositories.
- `nora retention-apply --yes` builds the same signer as the server so
  CLI-driven retention re-signs identically; plan/dry-run paths never
  regenerate and need no key.

Builds on the reconcile work (#847): the regeneration entry points it made
state-free are exactly what the post-deletion rebuild calls.

## Why

Hosted rpm/deb/raw repositories grow without bound (the storage-cost
formats: package streams, VM images), and were invisible to the one
mechanism meant to bound them. Deleting their versions safely requires
the index rebuild + re-sign, which is why this lands on top of the
reconcile groundwork rather than standalone.

## Checklist

- [x] Tests pass (`cargo test`) — 1618, incl. keep_last and age-only
      end-to-end over live test repos (packages + sidecars deleted,
      primary/Packages regenerated, signatures present), raw prefix
      grouping + root-file protection, name_glob targeting, dry-run
      no-op
- [x] New registry? See CONTRIBUTING.md checklist (not a new registry)


devitway (commented):
**Reviewed by reading the diff against `main` (153b892) and verifying each load-bearing claim in the source, adversarially (verify-contracts style).** Verdict: **TAKE** — the central safety claim holds; the notes below are follow-ups, not blockers.

## What I verified holds

- **"under the same per-repo publish lock the handlers use" — verified, exact string match.** Retention's post-delete rebuild locks `rpm/{repo}/repodata/repomd.xml` (rpm) / `deb/{repo}/Release` (deb). The upload/delete/reindex handlers lock `state.publish_lock(&repomd_key(&repo))` and `&release_key(&repo))`, and `repomd_key` = `rpm/{repo}/{REPODATA}/repomd.xml` with `REPODATA = "repodata"` (rpm.rs:58,68), `release_key` = `deb/{repo}/Release` (deb.rs:78). Same key → same `Arc<Mutex>` from the shared `publish_locks` map → a concurrent publish and a retention rebuild are genuinely serialized. No torn/half-signed index. This is the claim the whole PR rests on and it checks out.
- **Fail-safe collection throughout.** A sidecar that is unreadable, unparsable, or whose `href` diverges from its own key path is *skipped and kept*, never deleted (retention.rs collector). Retention can silently do nothing on bad metadata, but it can never delete the wrong thing.
- **raw root-file protection.** `raw/rootfile.bin` yields `(Some(top), None)` from `splitn(3,'/')` → `continue`; only `raw/{top}/{version}/…` is a collectable unit. Test asserts the root file survives.
- **Signer plumbing matches the stated contract.** Dry-run/plan pass `None` (no key needed); the real CLI path builds the same signer as the server (`build_signer`), the scheduler forwards `state.signer`. So a CLI retention pass re-signs identically to the server.
- **Ordering vs concurrent publish on the individual key deletes** reuses the existing per-key publish lock — unchanged, still correct.

## Notes / follow-ups (non-blocking)

1. **The "never advertises deleted packages" guarantee is best-effort on rebuild failure — worth stating as an accepted limitation.** Order is *delete packages+sidecars → rebuild+re-sign*. If the rebuild `Err`s, the deletions are already durable but the **old signed index stays**, now advertising packages whose payloads are gone (client 404s) until a publish or `-/reindex` heals it. The code handles this deliberately (loud error, points at `-/reindex`) — I'd just capture it explicitly (a one-line note in the retention docs / an accepted entry) so "retention can never leave a signed index advertising deleted packages" reads as "…once the rebuild succeeds; on rebuild failure the stale index persists until healed."

2. **Rebuild + re-sign fires for every matched repo even with zero deletions.** `regen.insert(...)` runs for any rpm/deb group whose rule matched, *before* and independent of `plans` being non-empty. A repo under a `keep_last`/`*` rule with nothing to evict gets its index rebuilt and **re-signed every cycle** — a fresh `InRelease`/`repomd.xml.asc` signature (new timestamp) on unchanged content, which churns client-side caches and spends a signing op per repo per run. Suggest gating the insert on an actual deletion (`if !plans.is_empty()`), so untouched repos aren't re-signed.

3. **The `name_glob` doc example for raw doesn't match.** `glob_match` is a plain glob (`*` spans any char, `/` not special). raw group names are a single segment `{top}` with no slash, so the documented `archive/*` (config/retention.rs) matches **nothing** — a user copying it gets silent keep-forever on exactly the unbounded-growth format this PR targets. The working pattern for raw is `archive` / `archive*`; the rpm/deb `myrepo/*` examples are fine (names carry the `/`). Worth fixing the example (and maybe a line clarifying raw names are single-segment).

4. **deb age = sidecar file mtime, not an upload timestamp** (deb metadata has none — `modified: None` → `stat(sidecar).modified`). If a future `-/reindex` rewrites a deb sidecar, that version's age resets and an `older_than_days` rule won't evict it. rpm is unaffected (uses `file_time` from inside the sidecar). Inherent to deb; just flagging that deb age-retention is approximate.

5. **Post-merge:** this adds `retention → signing` and `retention → registry::rpm/deb::regenerate_*` edges — the internal regression-map will want a re-seed once merged.

Nice piece of work — the reconcile groundwork (#847) paying off exactly as intended, and the lock discipline is right where it counts.

devitway:
**Landing-order flag** now that #852 (structured `dists/` layout) is open — this touches how retention here should group deb versions.

Retention collects deb versions grouped by `{repo}/{package}` (the sidecar's `package` field), with no distribution/component in the key. That's correct for the flat layout. But once #852 lands, the same package name can live in several distributions (`tree` in `jammy/main` **and** `focal/main`), and a `{repo}/{package}` group would fold them together — so `keep_last = N` would count across distributions and could evict a version that is the newest (or only) one in its own distribution.

The regeneration side self-heals (retention deletes package + sidecar and calls `regenerate_indexes`, which #852 makes structured-aware and prunes emptied dist trees), so the fix is purely in the *grouping*: group structured packages by `{repo}/{dist}/{component}/{package}` (flat stays `{repo}/{package}`), keyed off the sidecar `placement` that #852 introduces.

Since that grouping depends on #852's `placement` concept, suggested order: **land #852 first, then rebase this and make the retention grouping placement-aware.** Happy to re-review the rebased grouping. (Everything in my earlier review still stands for the flat path.)

devitway:
Thanks for carrying this through the rebase after #852 — the rpm/deb/raw coverage and per-repo rule targeting read well. One **blocker** before this can merge, plus a lower-severity note in the same function.

## Blocker: cross-distribution data loss for deb

Under the structured apt layout from #852, a package's `distribution`/`component` (its `Placement`) lives **only in the control sidecar** — it is not part of the storage key, and `regenerate_indexes` rebuilds each distribution's index from the surviving sidecars:

- `deb.rs` `package_key` → `deb/{repo}/{path}` — flat, no distribution (L78)
- `deb.rs` `Placement { distribution, component }` is a sidecar-only field (L122–129)
- `deb.rs` `regenerate_indexes` folds each *surviving* sidecar into `dists/{dist}/{component}/…` (L510+, L540)

Retention groups versions with no notion of placement:

- `collect_deb_versions` reads `package`/`version`/`arch`/`href`/`size` only — never `placement`/`distribution` (retention.rs L268–281)
- grouping key is `deb:{repo}/{package}` (retention.rs L337)
- `grep -nE 'placement|distribution|component' retention.rs` → 0 hits

So versions of the same package that belong to **different distributions** are pooled into a single `keep_last` scope. Since each distribution is an independent APT index, `keep_last=N` should count per distribution, not across the whole repo.

### Repro

Repo `myrepo`, structured layout:

- `PUT tree_1.8_amd64.deb?distribution=jammy` — jammy's only `tree`
- `PUT tree_2.0_amd64.deb?distribution=focal` — focal's only `tree`

Rule `{ registry: deb, name_glob: myrepo/*, keep_last: 1 }`.

`plan_deletions` sorts by mtime desc and deletes everything at index ≥ `keep_last`, so `tree_1.8` (jammy) is deleted. Its sidecar is removed, `regenerate_indexes` rebuilds, and **jammy loses `tree` entirely** — even though jammy had exactly one version and nothing in jammy exceeded `keep_last`. Any distribution that shares a package name with a newer sibling in another distribution is exposed.

### Suggested fix

Make the deb grouping key placement-aware: read `placement` from the sidecar and group by `deb:{repo}/{distribution}/{component}/{package}` (packages with no placement stay under the flat-root scope). Then `keep_last` is applied per distribution, matching how the index is actually rebuilt. Worth a regression test, e.g. `two_distributions_keep_last_evicts_sole_version` — PUT the two packages above, run `plan_deletions`, and assert jammy still resolves `tree`.

Flagging as a blocker rather than a follow-up because it's silent data loss on an ordinary multi-suite repo, which is exactly the case retention is meant to be safe for.

## Lower-severity note (same function)

`plan_deletions` breaks mtime ties with a lexical name compare — `b.name.cmp(&a.name)` (retention.rs L108) on `name = "{version}_{arch}"`. When mtimes tie — common for bulk/imported sidecars, since deb sidecars carry no upload time and fall back to the sidecar mtime (L333–335) — the lexical compare decides survivors, and `"1.9" > "1.10"` lexically, so the newer `1.10` is evicted. Not a blocker, but since you're already in this sort, a Debian-version-aware comparison here would avoid the surprise.

devitway:
Follow-up to the review above — I reproduced this end-to-end on a build of the PR head (`e219828`), so it's demonstrated silent data loss on a live instance, not just a code read.

**Setup** — local storage, deb registry, one retention rule:

```toml
[[retention.rules]]
registry = "deb"
keep_last = 1
```

**Two uploads of the same package into two distributions, one version each:**

```
PUT /deb/myrepo/pool/tree_1.8_amd64.deb?distribution=jammy&component=main  → 201
PUT /deb/myrepo/pool/tree_2.0_amd64.deb?distribution=focal&component=main  → 201
```

**Before retention** — each distribution resolves its package, and the two sidecars carry distinct placement:

```
GET dists/jammy/main/binary-amd64/Packages → 200  (Package: tree, Version: 1.8)
GET dists/focal/main/binary-amd64/Packages → 200  (Package: tree, Version: 2.0)

deb/myrepo/.nora-meta/pool/tree_1.8_amd64.deb.json → {"package":"tree","version":"1.8","placement":{"distribution":"jammy","component":"main"}}
deb/myrepo/.nora-meta/pool/tree_2.0_amd64.deb.json → {"package":"tree","version":"2.0","placement":{"distribution":"focal","component":"main"}}
```

**`nora retention-plan`** folds both into one group and targets jammy's only version:

```
[dry-run] would delete group=deb:myrepo/tree version=1.8_amd64 reason=beyond keep_last=1
```

**After `nora retention-apply --yes`:**

```
GET dists/jammy/main/binary-amd64/Packages → 404   ← jammy lost `tree` entirely
GET dists/focal/main/binary-amd64/Packages → 200   (tree 2.0 intact)
GET pool/tree_1.8_amd64.deb                → 404   (blob + sidecar deleted)
```

jammy held exactly one version of `tree` and nothing in jammy exceeded `keep_last=1`, yet it lost the package because retention counted `keep_last` across distributions — the placement-blind grouping (`deb:{repo}/{package}`) from the review. A `two_distributions_keep_last_evicts_sole_version` test reproduces it; grouping by `{repo}/{distribution}/{component}/{package}` (reading `placement` from the sidecar) fixes it.

jgrund:
Both addressed in 1c59f3f — thanks for the careful repro.

**Blocker:** `collect_deb_versions` now reads `placement` from the sidecar and groups structured packages as `deb:{repo}/{distribution}/{component}/{package}`; flat-root packages keep the old `deb:{repo}/{package}` scope, so `keep_last` counts per distribution — matching how `regenerate_indexes` rebuilds. Regression test `test_deb_keep_last_counts_per_distribution` covers your exact scenario (jammy + focal each holding `tree`, `keep_last: 1`): focal's sole version survives, and only jammy — which actually has two — gets trimmed.

**Tiebreak:** replaced the lexical compare with a numeric-aware one (`cmp_version_names`): digit runs compare numerically and `~` sorts before everything including end-of-string, so `1.10 > 1.9` and `1.0~rc1 < 1.0`. It's not the full Debian algorithm (no epoch/letter-vs-punct ordering) — the tiebreak is registry-agnostic and only decides among tied mtimes, so I kept it to the part that changes outcomes. Unit-tested including the tied-mtime `plan_deletions` path.

devitway:
The distribution/component fix looks right — thanks for turning it around so quickly. I re-verified it end to end, and there's one remaining grain of the same class (architecture) that I think is still a blocker. Details and a repro below.

### Confirmed fixed — cross-distribution eviction

`collect_deb_versions` now reads `placement` from the sidecar and groups structured packages per `{repo}/{distribution}/{component}/{package}` (`retention.rs:221`, `:292`). I checked the coupling on both ends: the writer serializes `Placement { distribution, component }` into the sidecar (`deb.rs:126`, `:310`), and the reader parses exactly those keys, so it can't silently drift to flat. `test_deb_keep_last_counts_per_distribution` drives the real upload path. `keep_last` now counts per distribution — the cross-distribution loss is gone. The numeric version tiebreak is a nice addition too.

### Remaining blocker — the same fold, one axis over (architecture)

The group key is `{repo}/{distribution}/{component}/{package}`, but `version` carries the arch (`"{version}_{arch}"`, `retention.rs:217`), and each `binary-{arch}/Packages` is an independent index — `regenerate_indexes` emits one per arch and filters `p.arch == arch || "all"` (`deb.rs:588-596`). So two architectures of the same package/version/distribution fold into **one** group, and `keep_last` can evict the sole version of one architecture. It's the cross-distribution failure one axis over, and it doesn't need mtime ties: `keep_last < (number of arches)` always drops an arch.

**Repro on a build of the PR head (`ee0937e`)** — one package, two real arches, one distribution:

```
PUT /deb/myrepo/pool/tree_1.8_amd64.deb?distribution=jammy&component=main   # Architecture: amd64
PUT /deb/myrepo/pool/tree_1.8_arm64.deb?distribution=jammy&component=main   # Architecture: arm64
# rule: { registry = "deb", keep_last = 1 }
nora retention-apply --yes
```

```
[dry-run] would delete group=deb:myrepo/jammy/main/tree version=1.8_amd64 reason=beyond keep_last=1
deleted group=deb:myrepo/jammy/main/tree version=1.8_amd64 ; indexes regenerated

GET dists/jammy/main/binary-amd64/Packages -> HTTP 404   # tree is gone for amd64
GET dists/jammy/main/binary-arm64/Packages -> HTTP 200   # tree 1.8 kept
```

The group the engine logs (`deb:myrepo/jammy/main/tree`) has no arch in it, so amd64 and arm64 share a single `keep_last` budget; amd64 loses its only version and its index is dropped as empty. An apt client on amd64 stops seeing the package. This slips through the tests because `build_test_deb_with` is amd64-only — every deb test is single-arch.

### Fix direction

Fold arch into the group key (`{repo}/{distribution}/{component}/{arch}/{package}`), symmetric to what you just did for dist/component. One subtlety: `arch = all` folds into every `binary-{arch}` index, so an `all` package is a single shared stream — it probably wants its own `all` group rather than being pooled under a concrete arch.

**rpm** has the same fold (`version = "{version}-{release}.{arch}"`, group `rpm:{repo}/{package}`). The client effect is softer there because rpm serves a single repodata instead of per-arch index files, but `keep_last` still collapses architectures, so it's worth the same grouping.

Happy to push up the failing regression test I used — two arches in one distribution, asserting both `binary-{arch}/Packages` keep the package — if that's helpful.

jgrund:
Right again — same fold, one axis over. Fixed in af8cce8:

- Group keys now carry the architecture: deb structured `{repo}/{distribution}/{component}/{arch}/{package}`, deb flat `{repo}/{arch}/{package}`, and rpm `{repo}/{arch}/{package}` (per your note that rpm collapses architectures the same way, just with softer client symptoms).
- `all`/`noarch` naturally form their own group — the arch comes from the sidecar as-is, so an `all` package never pools under a concrete architecture and its shared stream ages on its own budget.
- Regression tests for both formats: two architectures, one distribution, `keep_last = 1` — nothing planned, and both `binary-{arch}/Packages` (deb) / both payloads (rpm) survive. Your exact repro shape; the deb one goes through the real upload path with a real arm64 control paragraph, so the amd64-only fixture blind spot is closed too.

Thanks for the offer on the failing test — the shape you described is what I landed, but if yours covers an angle mine doesn't, happy to fold it in.

devitway (approved):
Verified end to end — `af8cce8` closes the remaining (architecture) blocker. Approving.

## Code

- Group key now carries the architecture on every path: deb structured `{repo}/{distribution}/{component}/{arch}/{package}`, deb flat `{repo}/{arch}/{package}`, and rpm `{repo}/{arch}/{package}` (`collect_sidecar_versions` in `retention.rs`). The arch is read from the sidecar as-is, so `all`/`noarch` packages form their own group instead of pooling under a concrete arch — which matches how `regenerate_indexes` emits one `binary-{arch}` index and filters `p.arch == arch || "all"`.
- Writer↔reader coupling holds: the deb handler parses the control `Architecture` field into the package record and serializes it into the sidecar (`deb.rs`), and `collect_deb_versions` reads exactly that field — the key can't silently drift back to a single-arch bucket. rpm is symmetric (arch from the package metadata).
- Both regression tests drive the real upload path with a real arm64 control paragraph, so the amd64-only fixture blind spot is closed as well.

## Live repro on a build of the PR head (`d60b315`)

Same polygon as the earlier repro, local storage, `keep_last = 1`:

- **Two arches, one distribution** (`tree_1.8_amd64` + `tree_1.8_arm64` in `jammy/main`): `retention-plan` reports 0 deletions and both `dists/jammy/main/binary-{amd64,arm64}/Packages` still serve `tree` (200/200). This is the exact case that returned 404 for amd64 on `ee0937e`.
- **Two distributions** (regression check for the first fix): `jammy` (1.8) + `focal` (2.0) both survive (200/200).
- **Positive control** — two versions of the same package in one arch+distribution: retention correctly evicts `1.8`, keeps `2.0`, and the `1.8` blob is gone (404). So the per-arch / per-distribution scoping isn't just suppressing eviction — `keep_last` still trims when a single index actually holds more than N. The dry-run even logs the group as `deb:myrepo/jammy/main/amd64/tree`, arch included.

No blockers remain. Thanks for turning both grains around so quickly.

ertime037 (approved):

PR #850, opened 2026-07-13
fix: gate the browse web UI on private (auth-enabled) deployments

jgrund:
## What

With auth enabled and `anonymous_read` off, the browse UI, its JSON
API (`/ui`, `/api/ui`), and the API docs (`/api-docs`) were served
without authentication — they enumerate every repository and package,
which is exactly what a private registry is there to keep private.

- The web surface now requires credentials unless `anonymous_read`
  (which already exposes the same names via the registry read APIs) or
  a new `auth.public_web_ui` (`NORA_AUTH_PUBLIC_WEB_UI`, default false)
  opens it. An unauthenticated request gets a `WWW-Authenticate: Basic`
  challenge so browsers prompt instead of showing a bare 401.
- `/metrics` moves behind its own `auth.public_metrics`
  (`NORA_AUTH_PUBLIC_METRICS`, default **true** — Prometheus scrapers
  rarely carry credentials, and metric labels name registry formats,
  not repositories); set false to gate it too.
- Health/readiness probes stay unconditionally public (a gated
  `/health` takes the whole deployment down behind a load balancer).
- Token-management pages remain always gated, as before.

`is_public_path` is split into the unconditional publics (probes,
token endpoints) and `is_web_surface`, which the middleware opens only
when configured — so the classification is explicit rather than a
wildcard that silently swept the UI into "public".

## Why

A private registry that authenticates its package APIs but serves an
open UI leaks its entire repository and package inventory to anyone who
loads `/ui`. Fail-closed is the right default for the web surface.

## Checklist

- [x] Tests pass (`cargo test`) — 1616, incl. the surface gated when
      private (with Basic challenge), probes/metrics still open,
      anonymous_read and public_web_ui each reopening it,
      public_metrics=false closing metrics, token pages staying gated
- [x] Updated CHANGELOG.md (Security)
- [x] New registry? See CONTRIBUTING.md checklist (n/a)


ertime037 (approved):

devitway (commented):
**Reviewed by reading the diff against `main` (153b892) and verifying each security-relevant claim in the source, adversarially.** Verdict: **TAKE** — a real information-disclosure fix, and the classification is now explicit rather than a wildcard. Details I checked:

## Verified

- **Fail-closed default.** `public_web_ui` is a bare `bool` with `#[serde(default)]`, so an operator who says nothing gets the surface **gated** — the safe direction. The behavior change (anonymous UI now needs `public_web_ui = true` or `anonymous_read`) is called out in both the CHANGELOG and `nora.env.example`.
- **Gating logic is correct and readable.** `is_public_path` is now *only* the unconditional publics (probes + the token endpoints that self-authenticate); `is_web_surface` is `/ui`, `/api/ui`, `/api-docs`; the middleware opens the surface *only* under `anonymous_read || public_web_ui`, and `/metrics` only under `public_metrics`. Splitting the old wildcard into two named predicates is the right call — it's why the UI stopped being swept into "public" by accident.
- **No token-page regression.** `is_web_surface` returns `false` for `/ui/tokens` and `/api/ui/tokens` *before* the `/ui` prefix match, so token-management pages stay gated regardless of `public_web_ui`. `/api/tokens*` remaining in `is_public_path` is unchanged from before (those endpoints do their own credential handling) — not introduced here.
- **Auth still runs ahead of the handlers** — this only changes which paths the middleware waves through, not the ordering.
- **`public_metrics = true` default is justified — I audited it, not just trusted the comment.** Every label set in `metrics.rs` is bounded: `registry` (the *format* — docker/maven/rpm/…), `method`, `status`, `decision`, `outcome`, `reason`, `result`, `operation`, `provider`. **None carries a repository, package, path, or artifact name.** So an open `/metrics` does not leak the inventory the UI gate is protecting. If a future metric ever gains a repo/package label, this default would need revisiting — worth a one-line comment at that label's definition, but nothing to change here.
- **Basic challenge** on an unauthenticated web-surface/`/metrics` request (`WWW-Authenticate: Basic`) so browsers prompt instead of rendering a bare 401 — good UX, and scoped to just those paths.

## Note

- The `/api/tokens*` "self-authenticating" property is load-bearing (they skip the auth middleware entirely) but pre-existing and untouched here — out of scope for this PR, just flagging that it's the one place where "public to the middleware" relies on the handler doing its own check.

Good, tightly-scoped security fix — fail-closed default and an explicit path taxonomy. Worth landing in the current patch line.

PR #852, opened 2026-07-13
feat: structured apt repository layout (dists/{distribution}/{component})

jgrund:
## What

The deb registry served only flat repositories (indexes at the repo root, sources line `deb <url>/deb/{repo} ./`). Flat is fine for small single-feed use, but the canonical apt layout — and the one existing repositories migrating to NORA are most likely already published in — is the structured `dists/` tree. This adds it:

- `PUT /deb/{repo}/{path}?distribution=<dist>[&component=<comp>]` places the package in the structured layout (`component` defaults to `main`); omitting `distribution` keeps today's flat behavior, and both layouts can coexist in one repository.
- Per distribution, regeneration writes `dists/{dist}/Release`, `InRelease`, `Release.gpg` and `{component}/binary-{arch}/Packages{,.gz}` for every component×architecture combination — empty ones included, so a Release never references a missing index. Arch-`all` packages fold into every concrete architecture index; a distribution holding only arch-`all` packages publishes `binary-all`.
- The upload path stays free-form: `Filename:` entries are repo-root-relative, so a `pool/` tree is conventional, not required.
- Placement is recorded in the package's control sidecar (old sidecars deserialize as flat), and the rebuild prunes what it no longer produces: root indexes once a repo is fully structured, a distribution's whole `dists/` tree when its last package is deleted, and stale signatures when signing is turned off (any of these left behind would keep advertising dead state to apt, signed).
- `dists/` is reserved against uploads and its indexes are served `Cache-Control: no-cache`, same as the flat reserved names. Distribution/component/architecture names are validated as single path-segment Release-list tokens before they reach a storage key.
- Reindex adopts sidecar-less packages into the flat layout: placement is upload-time intent and cannot be reconstructed from storage.

## Why

Structured is the layout modern apt tooling standardizes on (Debian, Ubuntu, and most third-party repos publish it; the Debian repository-format docs discourage flat for new repositories) and the one existing repositories are typically published in — supporting it makes NORA a drop-in target for them, while flat remains the zero-config default for simple feeds.

## Verification

`cargo test` — 1622 passing, including: dists-tree generation with hash-verified Release, arch-all folding, component×arch matrix (empty combos emitted), mixed flat+structured repos, dists-tree pruning on delete, placement-surviving reindex, and upload rejections.

Live against real apt (Debian trixie container, real dpkg-built packages from the Debian pool):

```
Get:1 http://…/deb/apt-e2e jammy InRelease [1783 B]
Get:2 http://…/deb/apt-e2e jammy/main arm64 Packages [512 B]
Get:3 http://…/deb/apt-e2e jammy/contrib arm64 Packages [478 B]
…
Setting up tree (2.2.1-1) ...
Setting up sl (5.02-1+b2) ...
```

- `apt update` verifies the per-dist InRelease via `signed-by` and fetches per-component/arch indexes.
- `apt install` from two components (main + contrib) succeeds.
- An unverifiable InRelease hard-fails `apt update` (`E: The repository … is not signed.`) — verification is enforced, not advisory.


devitway (commented):
**Reviewed by reading the diff against `main` and verifying each load-bearing claim in the source, adversarially (verify-contracts style).** Verdict: **TAKE** — the security-critical path handling is solid, the lock and consistency contracts are preserved, and the pruning is correctly scoped. Notes are follow-ups, not blockers.

## Verified

- **No path escape into storage keys.** `valid_release_token` is a strict whitelist — non-empty, ≤64, no leading `.`, chars `[A-Za-z0-9._+-]` only. No `/` is accepted, so a distribution/component/arch can never introduce a path separator; `!starts_with('.')` rejects `.`/`..`, and traversal needs a `/` anyway. It's applied to `distribution` and `component` in `resolve_placement`, and to `arch` (which comes from the control file) in `extract_record` before any `dists/{dist}/{comp}/binary-{arch}/` key is built. Uploads to `dists/` are rejected (`validate_package_path`: `path.starts_with("dists/")`). This is the one place I pushed hardest and it holds.
- **Lock discipline is preserved.** `regenerate_indexes` still documents "caller must hold the repo's publish lock", and the three callers (upload/delete/reindex) all take `state.publish_lock(&release_key(&repo))` = `deb/{repo}/Release`. The structured rebuild **and** the prune run under that single per-repo lock, so a concurrent publish can't interleave a half-built dists/ tree.
- **Pruning is scoped and self-healing.** After writing everything the rebuild produces (`desired`), it deletes only (a) RESERVED root names absent from `desired` and (b) keys under `deb/{repo}/dists/` absent from `desired`. Because uploads into `dists/` are forbidden, everything under `dists/` is generated — so pruning can't delete a package payload or a `.nora-meta` sidecar. "Drop root indexes once fully structured", "remove an emptied dist's whole tree", and "drop stale signatures when signing is off" all fall out of this one desired-set diff.
- **Consistency ordering intact.** Within `files`, each `Packages{,.gz}` is pushed before the `Release` that hashes it, and signatures last — a reader never sees a Release referencing bytes that aren't written yet. Every component×arch combination is emitted (empty included) so a Release never references a missing index; arch-`all` folds into each concrete arch, and an all-only dist publishes `binary-all`. Fail-closed on signing is unchanged.
- **Backward/forward-compatible sidecar.** `placement: Option<Placement>` with `#[serde(default, skip_serializing_if = "Option::is_none")]` — pre-structured sidecars deserialize as flat, and flat packages don't grow the sidecar. (This also keeps the retention collector reading old and new sidecars.)

## Notes (non-blocking)

1. **The prune isn't transactional with the writes.** If a `storage.delete` in the prune fails, the new indexes are already written but the call returns `Err` → the publish 500s to the client even though the package was stored and indexed; a leftover stale key (e.g. an old dist's Release) stays servable until the next successful rebuild prunes it. It's under the lock and self-heals, and delete-after-put failure is rare — just flagging that the window exists. A client retry is idempotent.
2. **The mid-rebuild hash-mismatch window is wider now.** With a full dists/ tree there are many more sequentially-written files, so a client holding an *older* Release has a longer window to hit a rewritten `Packages` (apt reports it and retries — same class as reprepro/aptly, and documented). If a hard guarantee is ever wanted, `Acquire-By-Hash` (by-hash index paths) is the standard closer — a reasonable future enhancement, out of scope here.
3. **`list deb/{repo}/dists/` on every publish/delete** — same full-list shape as the existing sidecar walk; fine at current scale, just noting it grows with the dist/component/arch matrix.
4. **Retention (#848) interaction.** Retention groups deb versions by `{repo}/{package}` with no distribution/component, so once this lands the same package name across two distributions would be conflated by `keep_last`. #852 itself is self-contained (no retention.rs change); this is a note for landing order — this first, then #848 rebased to group per-placement.

Live apt E2E (update + two-component install + hard-reject of an unverifiable InRelease) is exactly the right acceptance bar. Nice work.

ertime037 (approved):

Issue #583, opened 2026-06-01
OIDC namespace_scope is parsed and documented but never enforced (fail-open)


devitway:
## Summary

`OidcProvider.namespace_scope` is parsed and documented as a per-issuer access restriction, and `oidc.rs:11` advertises "Per-provider namespace scoping" as a security property — but the value is **never read or enforced** anywhere outside tests. Any token from a scoped issuer can write to any namespace its role allows. Fail-open: the config creates a false sense of security.

## Affected

- `nora-registry/src/config/auth.rs:235-238` (declaration + doc comment)
- `nora-registry/src/auth/oidc.rs` (`validate_token`, ~`115-219`)
- `nora-registry/src/auth/mod.rs` (`auth_middleware`, ~`268-292`)

## Impact — High

- Doc: `config/auth.rs:235-238` — *"which NORA namespaces this issuer can access. `["*"]` = all, `["github/*"]` = only repos under github/ prefix"*.
- `grep namespace_scope` over non-test `.rs` returns only the declaration in `config` and `vec!["*"]` in tests — **no read/compare site**.
- `validate_token` (`oidc.rs:206-218`) returns `OidcIdentity { provider, subject, issuer, role }` with no namespace field and no path check.
- `auth_middleware` admits the request based only on `role.can_write()`, never comparing the request path against `namespace_scope`.
- An admin who restricts a CI issuer to `namespace_scope=["github/*"]` expects its tokens to be unable to write elsewhere — there is no such restriction.

## Suggested fix

Either:
- **(a)** Enforce it: in `auth_middleware`, derive the target namespace from `request.uri().path()` and check it against `provider.namespace_scope` via `glob_match` (note the segment-boundary caveat in the related `glob_match` issue), returning 403 on mismatch; **or**
- **(b)** If the feature isn't ready, remove the field and its mention from the docs and from `oidc.rs:11` so operators aren't misled.

## Acceptance

- Integration test: a token from an issuer scoped `["github/*"]` receives 403 on `PUT /raw/other/...`.

Issue #853, opened 2026-07-14
Docs show docker login with OIDC JWT as Basic password, but Basic auth never tries OIDC validation


jgrund:
The OIDC docs suggest CI can authenticate docker with the workflow's OIDC JWT:

```yaml
- name: Push to NORA
  run: |
    echo "${{ steps.oidc.outputs.token }}" | \
      docker login registry.example.com -u oidc --password-stdin
```

That doesn't work against the current implementation. The Basic-auth path in `auth/mod.rs` tries htpasswd, then falls through to opaque API-token verification (`nra_…`, #736), then returns 401 — it never attempts OIDC JWT validation on the Basic password. Bearer OIDC works fine, but docker (and twine/maven) only speak Basic on this flow, so zero-secret docker push/pull from CI isn't possible as documented.

Suggested fix: in the Basic fallback, when htpasswd and opaque-token verification both fail and the password is JWT-shaped (two dots), run it through the OIDC validator with the same role/namespace-scope handling as the Bearer path. Alternatively, correct the docs until that exists.

Happy to send a PR for the Basic→OIDC fallback if that's the preferred direction.

devitway:
Confirmed — good catch, thanks. Let's do this in two steps.

**Now (simple):** I'll correct the OIDC docs so they don't advertise `docker login -u oidc --password-stdin <jwt>` until the Basic path actually supports it — a documented-but-broken zero-secret flow is worse than an honest "Bearer only for now."

**Next (correct):** the Basic→OIDC fallback you propose is the right fix — when htpasswd and opaque-token verification both miss and the password is JWT-shaped, run it through the OIDC validator with the same role/namespace-scope handling as the Bearer path. A PR would be very welcome.

One ask, since the Basic fallback is a **pre-auth surface**: keep the JWT-shape check strict (three base64url segments), and make a malformed/expired token fall through to a clean 401 — never a panic or a more-permissive branch. Fail-closed on any validator error.

PR #854, opened 2026-07-14
feat: per-rule namespace scope for OIDC role rules

jgrund:
## What

An `[[auth.oidc.providers.role_rules]]` entry may now set `namespace_scope` to narrow the provider's scope for identities matched by that rule:

```toml
[[auth.oidc.providers.role_rules]]
pattern = "repo:myorg/*:pull_request"
role = "write"
namespace_scope = ["ci-transport/**"]
```

Absent = inherit the provider's `namespace_scope` (existing behavior, and existing configs deserialize unchanged). Enforcement mode stays provider-level.

Also corrects the `role_rules` doc example in `config/auth.rs`, which showed a TOML map form that fails to parse — the real shape is an array of tables with `pattern`/`role`.

## Why

Namespace scoping (#583) is provider-wide, and provider lookup is by issuer, so a single issuer can't grant different blast radii to different workload identities. The concrete case: GitHub Actions pull-request builds (`sub = repo:org/x:pull_request`) need to stash run-scoped build artifacts, but giving the pull-request pattern a plain `write` role would let any PR build overwrite production repositories. With a rule-level scope, PR builds write only under a transport prefix while main/tag builds keep provider-wide write.

## Checklist

- [x] `match_role` returns the matched rule's scope; identity scope = rule scope or provider scope
- [x] Unit test: rule scope override picked, provider fallback preserved, first-match-wins unchanged
- [x] Integration test: PR-pattern JWT writes inside its rule scope (201) and is denied outside it (403) with provider scope `["*"]`
- [x] 1626 tests pass
- [x] CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 ## [Unreleased]
 
+### Added
+- **Per-rule namespace scope for OIDC role rules** — an `[[auth.oidc.providers.role_rules]]` entry may set `namespace_scope = ["ci-transport/**"]` to narrow the provider's scope for identities matched by that rule. Lets one issuer grant, e.g., pull-request CI builds write access confined to a transport prefix while main/tag builds keep the provider-wide scope. Absent = inherit the provider's `namespace_scope`; enforcement mode stays provider-level. Also corrects the config doc example for `role_rules`, which showed a map form that fails to parse (the real shape is an array of tables with `pattern`/`role`).
+
 ## [1.0.1] - 2026-07-13
 
 ### Security
--- a/nora-registry/src/auth/mod.rs
+++ b/nora-registry/src/auth/mod.rs
@@ -1522,10 +1522,17 @@ Jd74nq6dNCjpWG4drIsyhqX+
                         OidcRoleRule {
                             pattern: "repo:myorg/*:ref:refs/heads/main".to_string(),
                             role: "write".to_string(),
+                            namespace_scope: None,
+                        },
+                        OidcRoleRule {
+                            pattern: "repo:myorg/*:pull_request".to_string(),
+                            role: "write".to_string(),
+                            namespace_scope: Some(vec!["ci-transport/**".to_string()]),
                         },
                         OidcRoleRule {
                             pattern: "repo:myorg/*".to_string(),
                             role: "read".to_string(),
+                            namespace_scope: None,
                         },
                     ],
                 }],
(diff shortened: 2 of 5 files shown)


devitway:
Nice feature — per-identity blast-radius on a single issuer is a real gap, and the shape is clean. A few things I verified that look right, one design point I'd like changed before merge, and a forward-looking heads-up.

## Looks good

- **Backward compatible.** `#[serde(default, skip_serializing_if = "Option::is_none")]` on the new field means absent `namespace_scope` → `None` → inherits the provider scope; existing configs deserialize unchanged (`config/auth.rs`).
- **Empty scope is fail-closed.** `namespace_scope = []` → not unrestricted (no `*`) → `Scoped { patterns: [] }` → `any()` over an empty list is `false` → deny-all, and there's a test pinning it (`namespace.rs`, `empty_scope_is_fail_closed`). Good — that's the safe default for an empty allowlist.
- First-match-wins is preserved, enforcement mode stays provider-level, and the tests cover both the in-scope `201` and the out-of-scope `403`.

## Please change before merge: a rule scope should *only* narrow (intersect), not replace

Today the identity scope is a replacement:

```rust
// oidc.rs
namespace_scope: rule_scope.unwrap_or_else(|| provider.namespace_scope.clone()),
```

So a rule scope fully overrides the provider scope in either direction. A rule with `namespace_scope = ["*"]` promotes the identity to `Unrestricted` (`namespace.rs`: `scope.iter().any(|p| p == "*")` → `Unrestricted`) **even when the provider scope is narrow**. But the field is documented as *"Narrow the provider's `namespace_scope`"*, and the provider scope is documented as the provider's enforcement ceiling (`oidc.rs`, on `namespace_scope`). With replace, that ceiling isn't a ceiling — a rule can silently widen past it.

It's admin-authored config, so this isn't a privilege-escalation path (the `sub` is IdP-signed; a caller can't steer itself into a broader rule). The problem is the invariant: someone relying on the provider scope as a hard bound ("this issuer can never write outside X, whatever the rules say") is wrong under replace, and nothing makes that visible.

I'd rather make the narrowing structural than fix it in docs — a write should have to satisfy **both** the provider scope and, when present, the rule scope. Concretely: keep enforcing the provider scope, and if the rule sets a scope, run `enforce_namespace_scope` a second time against the rule authority; a write passes only if both allow. That keeps the provider scope a real ceiling, makes a rule that widens past it impossible to express, and matches the "narrow" contract the field already advertises. (If per-identity *widening* is actually intended, then the field isn't "narrow" and the provider scope isn't a ceiling — but that seems to cut against the motivating PR-build example, which wants a *narrower* scope than provider `["*"]`.)

## Heads-up: keep the Basic → OIDC path (#853) on this same code path

OIDC role + scope is resolved in exactly one place — `validate()` → `match_role` (`oidc.rs`, the only non-test caller). That's also why #853's JWT-as-Basic doesn't get a role today. If you land the Basic → OIDC fallback for #853, please route it through this same `validate()` so the rule scope is applied there too. A separate role-resolution path would drop the narrowing and fail open on the Basic transport — worth keeping in mind since you're touching both.

jgrund:
Agreed on all points — the ceiling invariant is the right call, and replace made it silently violable. Done in 9ddaa72 (conflicts with main also resolved):

**Narrowing is now structural.** `NamespaceAuthority::Scoped` holds a conjunction of pattern-sets and a write must match at least one pattern in *every* set — one set is the provider scope, a matched rule's scope adds a second. Each set still collapses independently: a set containing a bare `*` drops out of the conjunction (so `namespace_scope = ["*"]` on a rule is a no-op, not a promotion), and an empty set stays fail-closed. Widening past the provider ceiling is now impossible to express, per your suggestion. The identity carries `namespace_scope` (provider ceiling) and `rule_namespace_scope` separately, and the middleware builds the conjunction — `enforce_namespace_scope` runs both in one pass.

Tests pin: rule `["*"]` under a narrow provider stays scoped (unit + end-to-end 403 through the middleware), disjoint provider/rule scopes deny both sides (intersection, not union), all-star collapses to `Unrestricted`, empty rule scope locks out even under provider `["*"]`. Docs on `role_rules.namespace_scope` and the changelog now state the both-must-pass semantics.

**On the #853 heads-up:** noted, and agreed — if I pick that up, the Basic→OIDC fallback will route through `validate()` so rule scopes apply identically on that transport; role resolution stays single-path.

ertime037 (approved):

PR #859, opened 2026-07-16
fix: retention and GC schedulers run once at boot

jgrund:
Both schedulers skipped the interval's immediate first tick, deferring their first pass by a full `interval_secs`. A process that restarts more often than the interval therefore **never** runs retention or GC at all — we hit this in production: six rolling deploys in 48h meant the 24h retention schedule never fired once, while ~190 GiB of expired CI-transport artifacts accumulated under a rule that looked correctly configured.

- The first tick now fires immediately: one pass at boot, then every `interval_secs`.
- The boot pass **waits** on the shared cleanup lock instead of the usual skip-if-held — retention's and GC's boot passes fire at the same instant, and skip-if-held would silently postpone whichever loses the race by a whole interval again (the exact failure mode this fixes). Periodic passes keep skip-if-held semantics.
- Regression test: scheduler spawned with a 24h interval must delete a dominated version within seconds of boot.

Boot-time work is bounded: both passes are no-ops when there's nothing to collect, and the cleanup lock serializes them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ## [Unreleased]
 
 ### Fixed
+- **Retention and GC schedulers run once at boot** — both schedulers used to wait a full interval before their first pass, so a process restarting more often than the interval (rolling deployments, crash loops) never ran either of them, accumulating unbounded garbage while the schedule looked configured. The interval's first tick now fires immediately, and the boot pass waits on the shared cleanup lock (instead of skip-if-held) so retention and GC don't race each other out of their first run.
 - **Cargo sparse index now advertises `auth-required` on private deployments** — with auth enabled and `anonymous_read` off, `/cargo/index/config.json` sets `"auth-required": true` (RFC 3139) so cargo sends credentials on index and download requests. Previously cargo only authenticated the publish API, and every sparse-index fetch against a private instance failed with 401 before publish even started.
 
 ### Added
--- a/nora-registry/src/gc.rs
+++ b/nora-registry/src/gc.rs
@@ -835,8 +835,11 @@ pub fn spawn_gc_scheduler(
 ) -> tokio::task::JoinHandle<()> {
     tokio::spawn(async move {
         let mut interval = tokio::time::interval(std::time::Duration::from_secs(interval_secs));
-        // First tick fires immediately — skip it so GC doesn't run on startup
-        interval.tick().await;
+        // The interval's first tick fires immediately: GC runs once at boot,
+        // then every `interval_secs` — a process restarting more often than
+        // the interval otherwise never collects anything (see the matching
+        // note in `spawn_retention_scheduler`).
+        let mut boot_run = true;
 
         loop {
             // CANCEL-SAFETY: interval.tick() holds no state between polls.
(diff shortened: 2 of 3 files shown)


ertime037 (approved):

devitway (commented):
**Reviewed by reading the full scheduler loops, the `main.rs` call sites, and `run_retention`/`run_gc` against `main` (7f60f86), verifying each load-bearing claim in the source, adversarially.** Verdict: **TAKE** — a real bug (a process that restarts more often than the interval ran *neither* scheduler), a correct and minimal fix, and boot-time execution is safe where it matters. The two notes are follow-ups, not blockers.

## Verified

- **The bug is exactly as described.** The pre-loop `interval.tick().await` swallowed `tokio::time::interval`'s immediate first tick, so the first in-loop `interval.tick()` then waited a full `interval_secs`. A process whose lifetime is shorter than the interval therefore fired neither pass — confirmed in both `spawn_gc_scheduler` and `spawn_retention_scheduler`. Removing the swallow so the first tick runs at boot is the right fix.
- **The boot pass reads ground truth, not a cold cache — no data-loss hazard.** `run_retention` collects versions straight from `storage` (`collect_maven_versions(storage)` … `collect_deb_versions(storage)` …), never the in-memory `repo_index`. So a boot pass makes the *same* deletion decisions it would make a full interval later — boot timing is decision-equivalent, not a race against index warm-up. This is the property I pushed hardest on given the recent retention placement work (#848), and it holds.
- **The GC boot pass won't nuke in-flight startup writes.** `run_gc` skips any blob younger than `grace_secs` (`gc.rs:193`, `now.saturating_sub(meta.modified) < grace_secs`), so a boot GC racing a push that's still landing leaves the fresh blob alone.
- **`lock().await` on boot vs `try_lock` on periodic is the correct split.** Both boot passes fire at the same instant; if the boot path used `try_lock`, the loser would skip a whole interval — re-creating the exact bug being fixed. The `if boot_run { lock().await } else { try_lock() }` guard unifies to `Result<MutexGuard, TryLockError>` in both arms and compiles clean. Periodic passes keep skip-if-held, which is what you want (don't stack a second scan behind a running one).
- **Lock order is preserved.** The scheduler holds `cleanup_lock` across `run_*`, which take the per-path `publish_lock` *inside* — matching the documented `cleanup_lock → publish_lock` order (`main.rs:210`), no inversion introduced.
- **These are the first direct tests of either scheduler.** On `main` the only caller of both spawns is `main.rs`; this PR adds their first unit coverage. `..._runs_at_boot` (24h interval, must act within seconds) and `..._boot_run_waits_for_cleanup_lock` target the two right invariants, and neither is timing-fragile (10s deadlines, 50ms poll, real deletion as the signal).
- **No doc-drift.** The only "does not run on startup" text was the replaced comment; the CHANGELOG entry documents the new behavior.

## Notes (non-blocking)

**1. The boot `cleanup_lock.lock().await` sits outside the cancel `select!` — a graceful-shutdown latency regression on the first pass.** In both loops the boot acquire is a bare `.await`, not part of the `tokio::select! { cancel.cancelled() … interval.tick() … }`. So if a SIGTERM arrives while the *losing* scheduler is parked waiting for its sibling's boot pass, it won't observe the cancellation until it acquires the lock and then runs its own full `run_*` pass to completion (that pass isn't cancel-aware either). On a large store — the 190 GiB case from the description — that can overrun a 30s SIGTERM grace and get SIGKILL'd mid-pass. It's safe in the end (deletes are atomic/idempotent and the next boot re-runs), and the old `try_lock`-only path never *initiated* a pass after cancel — which is the regression. Optional tightening, keeps boot responsiveness:

```rust
let guard = if boot_run {
    boot_run = false;
    tokio::select! {
        _ = cancel.cancelled() => break,
        g = cleanup_lock.lock() => Ok(g),
    }
} else {
    cleanup_lock.try_lock()
};
```

**2. Under multi-replica this concentrates the pre-existing cross-pod cleanup race into the deploy window.** `cleanup_lock` is process-local (`main.rs:1608`, one `Arc<Mutex<()>>` per process), so it only serializes GC vs retention *within* a pod. Making both run at every boot means a rolling deploy of N replicas now triggers N concurrent retention+GC passes over the shared store, serialized only per-pod. This isn't introduced here and it's convergent (each pod computes the same delete set from storage; `grace_secs` guards in-flight writes) — but since the motivating incident is itself a rolling-deploy scenario, it's worth one line in the docs that the boot pass runs per-pod and cross-pod serialization needs a distributed lock (future work), the same single-writer framing already applied to the signing key in #841. For single-writer / RWO deployments this is a non-issue.

The production repro (6 deploys in 48h, a 24h schedule that never fired, 190 GiB accumulated) plus the two boot-behavior tests are the right acceptance bar. Tightly-scoped, well-motivated fix — worth landing.

PR #864, opened 2026-07-21
feat: pull-through proxy and offline mirror for rpm and deb

jgrund:
## What

Adds the proxy mode to the last two hosted-only formats, plus offline pre-fetch:

- **Server pull-through**: `[rpm.proxies]` / `[deb.proxies]` map a local repo name to one upstream repo URL (string or `{ url, auth }`; env `NORA_{RPM,DEB}_PROXIES="repo=url|auth,…"`). Proxied repos are read-only (publish/delete/reindex → 409). Upstream metadata (`repodata/`, `dists/`, the upstream's own signatures/keys) is served verbatim within `metadata_ttl` (default 300 s; non-positive revalidates every pull); packages (`.rpm`/`.drpm`, `.deb`/`.udeb`) cache immutably; when the upstream is down the stale cache is served with `x-nora-stale: true`.
- **Shared flow**: one `repo_proxy_download` in `registry/mod.rs` used by both handlers, on top of the existing `proxy_fetch` / circuit breaker / `spawn_cache` machinery. Upstream hosts register with the leak detector; the digest quarantine gates proxied packages via `[curation.rpm]` / `[curation.deb]` (never the mutable indexes), so rpm/deb keep security parity with the other proxy formats.
- **Offline mirror**: `nora mirror rpm --repo <name> [--arch x86_64,noarch]` and `nora mirror deb --repo <name> [--dist bookworm] [--component main] [--arch amd64]` enumerate the upstream package list *through* the proxy (repomd.xml → primary.xml; Release → Packages in `.gz`/`.xz`/plain) and pre-fetch every package, HEAD-skipping what the cache already holds. After a run, dnf/apt clients work with the upstream unreachable. No new dependencies — index parsing is a string scan over machine-generated XML/control paragraphs, decompression via the already-present flate2/lzma-rs/ruzstd.

Design notes: the upstream map is per-repo (rpm/deb have no global coordinate space) and single-upstream-per-repo — mirrors of the same distro repo lag each other, and mixing them within a TTL window can serve a repomd.xml whose referenced blobs come from a different sync generation. Dependency-confusion gating doesn't apply (a repo is either proxied or hosted by config, never falling through), which is why the proxy path carries the quarantine but not the namespace-isolation checks.

## Why

RPM and DEB were the only formats without an upstream proxy (COMPAT listed "hosted only"), so a NORA in front of Fedora/EPEL/Debian/Ubuntu repos couldn't cache them, and air-gapped sites couldn't mirror them. Follows the phasing noted when hosting shipped (#128).

## Checklist

- [x] Tests pass (`cargo test`) — 13 new server proxy tests (cache-hit without refetch, ttl=0 revalidation, stale serve, quarantine hold, write rejection, flat `./` strip) + 2 wiremock end-to-end mirror runs + parser/env unit tests
- [ ] New registry? See CONTRIBUTING.md checklist — not a new registry

https://claude.ai/code/session_018WQbaFrgw82c7KCcbhpRbQ


ertime037 (approved):

devitway:
Reviewed the full diff and traced the shared machinery on a local checkout of the PR branch — approving. The read-only enforcement, quarantine gating, graceful degrade and stale-fallback all check out, and the test coverage (XML/Release parsers + wiremock integration for fetch/cache/stale/ttl/quarantine/read-only) is good.

**Verified**
- Read-only proxied repos: all of upload/delete/reindex return 409 for both formats (`proxied_repo_conflict`), with tests.
- `repo_proxy_download` graceful-degrade: `storage.get().ok()`; a failed `stat` makes `mutable_ref_fresh` return false → revalidate upstream; upstream down → stale copy with `x-nora-stale: true`. No hard-fail on cache I/O.
- Quarantine gates immutable packages on all three serve paths (fresh-cache-hit, upstream-fetch, stale-fallback), never the mutable metadata — correct, since metadata digests rotate every upstream sync.
- `validate_storage_key` + the `META_DIR` (`.nora-meta`) guard run before any storage access on the proxy path.
- Auth handling is clean: `ProtectedString` + `skip_serializing`, passed out-of-band from the URL, so it can't leak into logs.

**One follow-up (non-blocking): cache eviction.**
Proxy packages are cached via `spawn_cache_immutable` with no sidecar, so retention (sidecar-based) never groups them and GC lists rpm/deb under `uncovered`. Combined with `nora mirror` bulk-fetching an entire upstream, the cache can grow with no eviction path. It's consistent with the other proxy formats, so nothing this PR introduces — but `mirror` makes it acute. I'll file a follow-up for a size/LRU-bounded proxy cache plus a COMPAT note; not a blocker here.

**Nits**
- `mirror/mod.rs` `warm_repo_paths`: the "HEAD 200, local-only check" comment is a bit misleading — `client.head()` routes to `download()` (axum `get()` also serves HEAD), so for an un-cached path it runs the full proxy flow and hits upstream (caching as a side effect). Works fine, but the byte counter also under-counts paths warmed via HEAD (bytes are only tallied on the GET branch).
- `digest_quarantine.rs` `resolve_global` doc ("the proxy gate uses the global setting for now") is now stale — `repo_proxy_download` passes the per-registry `curation.{rpm,deb}` override.
- `parse_repo_proxies_env`: `repo=url|auth` split on `,` then `=` then `|` — an auth value containing `,` would break record splitting. Fine for `user:pass`/tokens; worth a note on the env format.
- Edge: a repo previously used as hosted (with data on disk), then added to `[*.proxies]`, will serve its old hosted packages as immutable proxy cache under the same `package_key`. Might be worth a startup warning.

PR #885, opened 2026-07-28
fix(maven): keep server-generated artifact metadata authoritative

### LINKED ISSUE #886: bug(maven): proxy metadata merge writes maven-metadata.xml + checksums without publish_lock
### Summary

`merge_and_cache_proxy_metadata` (added in #885) writes `maven/<g>/<a>/maven-metadata.xml` and its four checksum sidecars on the **download** path without holding the artifact's `publish_lock`. The version-upload path regenerates the same key **under** `publish_lock` (`metadata_lock_key`). Because the two paths do not share the lock, a proxy metadata GET and a concurrent version PUT on the same artifact (or two concurrent GETs) serialize on nothing.

This weakens the invariant *"`publish_lock` serializes all writes to the same artifact path"* for the metadata document, and can leave the stored `.sha1`/`.md5`/… corresponding to different bytes than the stored `.xml`: the document and its four checksums are five independent, non-atomic `put`s. A client that fetches `maven-metadata.xml` and then `maven-metadata.xml.sha1` inside that window sees a checksum mismatch.

### Evidence

- **upload takes the lock:** `nora-registry/src/registry/maven.rs` — `let lock = state.publish_lock(&metadata_lock_key); let _guard = lock.lock().await;` then `update_artifact_metadata(...)` (which writes the doc + 4 checksums).
- **download merge does not:** `merge_and_cache_proxy_metadata` → `state.storage.put(&key, &data)` + `compute_and_store_checksums(...)` with no `publish_lock`.
- **Reachable** only when an artifact is both proxied and hosted — the exact hosted/proxy collision #885 targets.

### Impact

- Single-instance: transient, self-healing checksum-mismatch window on `maven-metadata.xml` under concurrent GET + PUT. No permanent lost version (the version list is scan-regenerated from storage on the next write).
- Consistency, not data loss. Severity: medium.

### Fix

Acquire `state.publish_lock(&doc_key)` in `merge_and_cache_proxy_metadata` around the read → merge → write → checksums cycle — the same key the upload path locks, so the two serialize and the document + its checksums are written as one critical section.

Depends on / folds into #885.

ivaseeq:
Maven clients may upload stale artifact-level `maven-metadata.xml` after concurrent deploys, while a proxy refresh may replace hosted versions with the upstream document.

This change keeps Nora-generated artifact metadata authoritative during deploys and merges stored hosted versions into refreshed upstream metadata. Metadata checksums are generated from the merged document.

Testing:

- Added regression coverage for stale metadata uploads, checksums, and both hosted/proxy request orders.
- Verified concurrent deploys, RELEASE/SNAPSHOT resolution, proxy caching, hosted/public coordinate collisions, and failover with S3-backed single- and multi-replica deployments.

---

Also serializes the proxy-side metadata merge under `publish_lock`, so the metadata document and its checksum sidecars are written as one critical section and stay mutually consistent under concurrent fetch and publish.

Closes #886.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 ## [Unreleased]
 
+### Fixed
+- **Maven keeps server-generated artifact metadata authoritative** — a Maven client that re-uploads a stale artifact-level `maven-metadata.xml` after a concurrent deploy no longer overwrites the version list NORA generates: an uploaded artifact-level metadata document (and its checksums) is recognized by its shape and dropped, while version-level (SNAPSHOT) and group-level (plugin) metadata are still stored verbatim. On a proxy refresh, locally hosted versions are merged into the refreshed upstream document instead of being replaced by it, and the `.md5`/`.sha1`/`.sha256`/`.sha512` sidecars are recomputed from the merged document. The proxy-side merge runs under the same `publish_lock` as the upload-side regeneration, so the document and its checksums are written as one critical section and stay mutually consistent under concurrent fetch and publish (#886).
+
 ## [1.1.0] - 2026-07-26
 
 ### Fixed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2351,6 +2351,7 @@ dependencies = [
  "pgp",
  "prometheus",
  "proptest",
+ "quick-xml 0.41.0",
  "rand 0.8.6",
  "reqwest",
  "rpm",
@@ -2550,7 +2551,7 @@ dependencies = [
  "md-5 0.10.6",
  "parking_lot",
  "percent-encoding",
- "quick-xml",
+ "quick-xml 0.39.3",
  "rand 0.10.1",
  "reqwest",
  "ring",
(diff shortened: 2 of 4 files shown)


ertime037 (approved):
Thanks @ivaseeq — the shape-based classification (drop artifact-level client metadata, keep version/group verbatim) and the hosted+proxy merge are the right model, and the test coverage is thorough.

On review I found one concurrency gap in the new proxy path: merge_and_cache_proxy_metadata wrote maven-metadata.xml and its four checksum sidecars without the publish_lock the upload-side regeneration holds, so a concurrent proxy GET and version PUT could leave a stored .sha1 that no longer matches the stored .xml (transient and self-healing, but a client fetching both inside that window sees a mismatch). Tracked as #886 and folded the fix into this PR (2242efb): the merge now runs under the same lock on the document key, plus a CHANGELOG entry. fmt, clippy -D warnings and the 67 maven tests are green on the combined branch.

Approving. Heads-up: main requires signed commits, so this lands via the `reland` label as one web-flow-signed commit — your authorship is kept through Co-authored-by.

devitway:
🤖 Relanded as a signed commit: https://github.com/getnora-io/nora/pull/887 (thanks @ivaseeq — full authorship credited via `Co-authored-by`). This PR will be closed once the reland merges.

devitway:
Landed on `main` via signed reland #887 (commit 0644178). Thanks @ivaseeq — full fix (shape-based metadata authority + hosted/proxy merge + the follow-up publish_lock serialization from #886) is in. Closing in favor of the reland.

PR #889, opened 2026-08-01
feat(registry): add Nexus-compatible named Maven and npm registries

ivaseeq:
## What changed

- adds named Maven hosted, proxy, and group repositories with Nexus-compatible routes and policies
- adds separate npm hosted, proxy-cache, and ordered group repositories
- keeps npm hosted manifests, tarballs, tags, deprecations, package fields, and rebuildable packument cache separate from proxy state
- materializes hosted npm packuments so warm metadata reads use one object-store GET instead of one GET per package version
- invalidates the derived packument cache under the package mutation lock for publish, dist-tag, deprecation, and retention changes
- adds conditional immutable writes, retention, GC, indexing, metrics, UI, auth, and OpenAPI support for the named layouts

## Why

This enables a fresh Nora installation to replace the developer-facing Maven and npm flows currently served by Nexus. Clients retain their standard Maven and npm commands; deployments change only registry endpoints and authentication material.

The packument cache fixes a migration and developer-flow failure observed with packages containing thousands of versions: every metadata request previously fanned out into thousands of S3 reads and exceeded the ingress timeout. The cache is derived and rebuildable; authoritative hosted state remains split by responsibility.

The target topology is a singleton writer. Migration from older Nora object layouts and active-active mutable metadata coordination are intentionally outside this change.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy -p nora-registry --all-targets -- -D warnings`
- library tests: 63 passed
- binary tests: 1,797 passed, 1 ignored
- documentation tests: 6 passed, 1 ignored
- `git diff --check`
- targeted tests cover cold/warm packument reads, per-route tarball URL rewriting, mutation invalidation, fail-closed invalidation, retention, and proxy-cache isolation
- combined Maven/npm acceptance and live Nexus-to-Nora migration are tracked separately against the digest built from this source


devitway:
@ivaseeq — thank you for this, it is careful work. Sorry that we can't take all of it yet: the named-repository model moves enough of the routing and the storage layout that we are not ready to commit to that shape, and we are still thinking it through.

What we could take, we did. The packument rebuild and the abbreviated `install-v1` packument both landed in #958, and in #962 npm retention stopped advertising versions whose tarballs it had already deleted — that last one we found while running your branch. Your co-authorship is on the first two.

We did run it before deciding, against S3 with 500 versions, and a couple of measurements from that might be worth having regardless of what happens here — the one that would matter most to you is that data written by a released NORA reads back as 404 under the new layout.

Still thinking about the rest.

PR #892, opened 2026-08-05
fix(storage): size object_store timeouts for client-paced streaming

jgrund:
## Symptom

Any download from an object-store backend (GCS/S3 mode) that cannot complete within ~180 seconds is reset mid-stream with an HTTP/2 RST, deterministically. Slow WAN clients pulling multi-GB artifacts hit this every attempt; fast clients never see it. Server logs show repeated `object_store::client::get` — "Encountered error while reading response body … Retrying in 0.1s" — before the reset.

## Cause

Downloads stream store → server → client under backpressure (`Body::from_stream` in the raw path and equivalents), so a slow download client paces the store read. Two `object_store` defaults interact:

- `ClientOptions` request timeout (**30s**) has total-request semantics — it covers the streamed body, so a client-paced GET is killed every 30s;
- each kill is transparently resumed with a ranged re-request (`client/get.rs` `retry_stream`), but every resume draws on one `RetryConfig::retry_timeout` deadline (**180s**) that starts at the *first* request (`RetryContext::exhausted`).

At 180s of wall clock the next resume is refused, the stream errors, and the response body aborts.

## Fix

Raise the per-attempt timeout to 1h and the retry deadline to 2h on both the GCS and S3 builders (`storage/object.rs`). On the GCS path the timeout is merged via `with_config` rather than `with_client_options` so env-derived client options (`GOOGLE_PROXY_URL`, …) survive, and an operator's explicit `GOOGLE_TIMEOUT` still wins.

Verified in production (GKE, GCS mode): with the equivalent `GOOGLE_TIMEOUT=3600s` an operator-side workaround, slow transfers complete; without it, curl dies at ~185s with exit 92 (`--limit-rate 3M` on a 1.66 GB object reproduces in ~3 min).

`cargo check -p nora-registry` clean.

## Related

While tracing this: the raw registry ignores HTTP `Range` headers (200 + full body). Range support would let interrupted CI pulls resume instead of restarting, but it interacts with the streaming integrity-pin gate (`VerifyingStream` can only verify a complete body), so it needs a design decision — happy to file separately.

https://claude.ai/code/session_01JyQLELHdpBGp8AjUTS94Ej

### DIFF
--- a/nora-registry/src/storage/object.rs
+++ b/nora-registry/src/storage/object.rs
@@ -86,9 +86,10 @@ impl ObjectStorage {
             .with_endpoint(url)
             .with_bucket_name(bucket)
             .with_region(region)
-            .with_allow_http(allow_http)
+            // One combined ClientOptions: with_client_options REPLACES the
+            // options with_allow_http would have accumulated.
+            .with_client_options(streaming_client_options().with_allow_http(allow_http))
             .with_virtual_hosted_style_request(virtual_hosted)
-            .with_client_options(streaming_client_options())
             .with_retry(streaming_retry_config());
 
         match (access_key, secret_key) {
--- a/nora-registry/src/storage/object.rs
+++ b/nora-registry/src/storage/object.rs
@@ -13,6 +13,26 @@ use tokio::io::{AsyncRead, AsyncReadExt};
 
 use super::{FileMeta, Result, StorageBackend, StorageError};
 
+/// Downloads stream store -> server -> client under backpressure, so a slow
+/// download client paces the store read. object_store's request timeout covers
+/// the whole body of an attempt, and its retry deadline runs from the FIRST
+/// request while gating every transparent ranged resume of a streamed body —
+/// with the defaults (30 s attempt, 180 s deadline) any transfer slower than
+/// object-size / 180 s is reset mid-stream. Size both for the slowest
+/// tolerated download instead.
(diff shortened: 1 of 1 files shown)


devitway:
Approve. The diagnosis is precise and matches object_store's total-request timeout semantics; the GCS `with_config` path correctly preserves env-derived options and yields to an operator's `GOOGLE_TIMEOUT`. fmt/clippy/1784 tests green locally.

One accepted-limitation note before merge: `with_timeout(3600s)` is set on the client builder, so it's **global to every object_store op** on that backend, not just streamed bodies — a hung small metadata GET/PUT now fails in ~1h instead of 30s. That's the right trade (object_store exposes a single total-request timeout; connect vs body-idle can't be split), but please record it as an accepted limitation (one line in the fn doc or COMPAT) so the `streaming_` name doesn't imply a scoping that isn't there. Minor: 1h is a fixed ceiling, so a transfer slower than object-size/1h still resets — fine for the target, worth a comment.

ertime037 (approved):

devitway:
🤖 Relanded as a signed commit: https://github.com/getnora-io/nora/pull/894 (thanks @jgrund — full authorship credited via `Co-authored-by`). This PR will be closed once the reland merges.

jgrund:
All of this landed on main as 72e349f via #894 (reland/pr-892) — the diff against main is now empty. Closing.

PR #893, opened 2026-08-05
feat(registry): resumable downloads (Range/206) for all formats

jgrund:
Extend the Range support docker blob GET gained in #657 to every artifact payload: maven, npm, pypi, gems, cargo, go, nuget, terraform, conan, deb, rpm, ansible, pub, and raw.

A single-range request is served from the storage backend's native ranged read (local file seek, S3/GCS ranged GET) through a new shared helper in registry/range.rs; the parser, its Kani proof, and the proptest move there from docker.rs. A resume at end-of-file now gets 416 + "Content-Range: bytes */{size}" instead of a full re-download, a failed ranged read falls back to the full 200 instead of a 500, and full-200 artifact responses advertise "Accept-Ranges: bytes".

Mutable content (maven-metadata.xml, packuments, indexes, dists/, repodata/) neither advertises nor honors ranges — a resumed range across a rewrite would splice two generations. raw, the one overwritable format, honors If-Range against its pin ETag. A ranged serve carries no server-side integrity check (a partial body cannot be re-hashed; the client's own checksum covers it — the #657 precedent), and where the digest-quarantine gate needs the whole object a range request under an active quarantine policy falls back to the gated full response rather than bypassing it.

Claude-Session: https://claude.ai/code/session_015F6q43HAb8k2bALqS4zLdu

## Checklist

- [ ] Tests pass (`cargo test`)
- [ ] No new clippy warnings (`cargo clippy -- -D warnings`)
- [ ] Updated CHANGELOG.md (if user-facing change)

---
**Stacked on #895** — this branch is rebased onto its head, so the diff shows #895's commit until it merges; after that this PR shows only the Range/206 work.


devitway (dismissed):
Approving — this is excellent work.

The `range.rs` parser is Kani-proven bounds-safe over the whole `u64` space (`start ≤ end < size`, no overflow), the mutable-content guard is enforced per format — explicit path/type checks on maven/npm/nuget/deb/rpm/gems plus raw's `If-Range` against the pin ETag, handler-scoping elsewhere, backed by negative tests (`test_go_mod_ignores_range`, the terraform provider/module pair, the deb index assertion) — the digest-quarantine fallback stands down correctly, and fmt/clippy/1784 tests are green locally. Consolidating the docker #657 range path into one shared helper is the right call.

One thing to add before merge: a one-line note in `SECURITY.md` / `COMPAT.md` that 206 (ranged) serves carry **no server-side integrity verification** — a partial body can't be re-hashed, so the client's own checksum (lockfile hash, content digest, `go.sum`) is the compensating control (the #657 precedent, now generalized to all formats), and under an active digest-quarantine policy the range serve stands down to the gated full response. The code comments already say this clearly; it just deserves to be user-visible in the security-posture docs. Could you add it to this PR?

I'll land the accompanying review-tooling update on our side at merge — nothing else needed from you there.

ertime037 (approved):

PR #895, opened 2026-08-06
fix(docker): return OCI-conformant 429 for concurrent-upload limit

jgrund:
## Problem

The concurrent-upload ceiling returned `429` with a plain-text body (`Too many concurrent uploads`). podman/skopeo parse registry errors through docker/distribution errcode: an OCI error JSON with code `TOOMANYREQUESTS` is classified retryable by containers/common retry logic, but a plain-text body becomes `UnexpectedHTTPStatusError` and aborts the whole push on first 429 instead of backing off. Hit in real CI fan-out pushing to nora.

## Fix

One shared helper `too_many_uploads()` replacing all three plain-text sites (session-limit check in upload start, `InFlightGuard` checks in streaming PATCH/PUT):

- `Content-Type: application/json`, body `{"errors":[{"code":"TOOMANYREQUESTS","message":"too many concurrent uploads","detail":{"limit":<max_upload_sessions()>}}]}`
- `Retry-After: 5`

Limit semantics unchanged — response shape only.

## Test

`test_upload_session_limit_returns_oci_429` pre-fills the session map to the ceiling (no env mutation, parallel-safe) and asserts 429, `TOOMANYREQUESTS` code, `Retry-After`, and `detail.limit`.

`cargo test -p nora-registry`: 1698 passed; 0 failed; 1 ignored.

https://claude.ai/code/session_01V6Vgp1PEVp8zkhAZtZ6Bz3

### DIFF
--- a/nora-registry/src/registry/docker.rs
+++ b/nora-registry/src/registry/docker.rs
@@ -481,6 +481,28 @@ fn max_upload_sessions() -> usize {
         .unwrap_or(DEFAULT_MAX_UPLOAD_SESSIONS)
 }
 
+/// OCI-conformant 429 for the concurrent-upload ceiling. containers/common
+/// retry logic only backs off on errcode TOOMANYREQUESTS in a JSON body; a
+/// plain-text 429 aborts the client push outright.
+fn too_many_uploads() -> Response {
+    let body = json!({
+        "errors": [{
+            "code": "TOOMANYREQUESTS",
+            "message": "too many concurrent uploads",
+            "detail": { "limit": max_upload_sessions() }
+        }]
+    });
+    (
+        StatusCode::TOO_MANY_REQUESTS,
+        [
+            (header::RETRY_AFTER, "5"),
+            (header::CONTENT_TYPE, "application/json"),
+        ],
+        body.to_string(),
+    )
+        .into_response()
+}
+
 /// Read max session size from env (in MB) or use default
 fn max_session_size() -> usize {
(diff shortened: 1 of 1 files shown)


devitway:
Approving — clean, correct, and OCI-conformant. Verified against the code and the spec.

**Completeness:** all three plain-text rejection sites now route through the shared `too_many_uploads()` helper — the session-limit check in `start_upload`, and the `InFlightGuard` misses in `patch_blob` (PATCH) and `upload_blob` (PUT). Grepped the file at the PR head: no other plain-text 429 remains (`TOO_MANY_REQUESTS` now appears only in the helper and the test).

**OCI conformance** (checked against distribution-spec `spec.md`):
- Envelope `{"errors":[{"code","message","detail"}]}` matches the spec's error format (§ Error Codes).
- `TOOMANYREQUESTS` is the spec's `code-14`, and it's a valid `code` (uppercase + underscores, per the MUST on the `code` field).
- `detail: {"limit": …}` is allowed — the spec says `detail` MAY be arbitrary JSON the client can use to resolve the issue.
- `429` + `Retry-After` satisfies the spec's SHOULD for 429 responses.

The test covers status, `Content-Type: application/json`, `Retry-After`, the `TOOMANYREQUESTS` code, and `detail.limit`, and it's parallel-safe (pre-fills the session map, no env mutation). Limit semantics are untouched — response shape only.

Nice catch on the real failure mode: containers/common only backs off on the JSON errcode, so a plain-text 429 aborts the whole push instead of retrying. LGTM.

ertime037 (approved):

Issue #902, opened 2026-08-09
`min_release_age` blocks all packages as `0h old` under `trust_upstream_dates=true` on v1.1.0


patsevanton:
## Summary

On NORA **v1.1.0** (image `ghcr.io/getnora-io/nora:1.1.0`, helm chart `nora-0.4.4`), the
`min_release_age` curation filter reports **every** proxied artifact as `0h old`, regardless of its
real upstream publish date — even packages published years ago. As a result, in `mode=enforce` the
filter blocks **all** downloads (including provably old packages like `lodash@4.17.21` from 2021),
making the registry unusable. In `mode=audit` it produces noisy false-positive `[AUDIT] Download
would be blocked` log lines for every package.

`server.trust_upstream_dates = true` is set, which per the CHANGELOG (0.9.5, #748) is supposed to
seed the quarantine `first_seen` from the trusted upstream release date so that an old artifact
matures immediately. On v1.1.0 this does not happen: the upstream date is never used for the
release-age decision, and every package is treated as "new to this mirror".

## Environment

- NORA version: `1.1.0` (image `ghcr.io/getnora-io/nora:1.1.0`)
- Helm chart: `nora-0.4.4`
- Storage backend: S3 (Yandex Object Storage)
- Registries tested: `npm`, `pypi`, `cargo`, `go`
- Deploy: Kubernetes (Yandex Managed K8s), ingress-nginx + cert-manager

## Configuration

`helm-values.yaml` (relevant parts):

```yaml
config:
  server:
    public_url: "https://nora.111.88.255.129.sslip.io"
    trust_upstream_dates: true
  curation:
    mode: "enforce"          # also reproduced with "audit"
    on_failure: "closed"
    min_release_age: "7d"
    blocklist:
      existingConfigMap: "nora-curation-blocklist"
      key: "blocklist.json"
      mountPath: "/etc/nora-curation/blocklist"

extraEnv:
  - name: NORA_CURATION_QUARANTINE
    value: "enforce"          # also reproduced with "observe"
```

> Note: helm chart 0.4.4 does **not** render `quarantine` under `[curation]` in `config.toml`, so
> `NORA_CURATION_QUARANTINE` must be supplied via `extraEnv`. Without it NORA refuses to start
> (config guard from #741). This is a separate chart limitation, not the subject of this report.

## Reproduction

1. Configure NORA as above (`mode=enforce`, `min_release_age=7d`, `trust_upstream_dates=true`,
[...]

Issue #750, opened 2026-06-19
curation: min_release_age provides no release-age freshness in any config (regression from #742)


devitway:
## Summary

After #742 repositioned `min_release_age` to defer to quarantine, **`min_release_age` + `server.trust_upstream_dates` no longer provides release-age freshness in any configuration**. Reported by @pyther in #748 (PyPI), but the behavior is config/curation-level and identical for every proxy registry.

## Reproduction (an OLD package, e.g. `six==1.16.0` (2021) on PyPI, `is-number@7.0.0` (2018) on npm)

| config | result on the old package | mechanism |
|---|---|---|
| `quarantine=off` + `min_release_age` | **rejected at startup** | config guard: min-age on an enabled proxy requires quarantine; `trust_upstream_dates` is explicitly *"not a substitute"* |
| `quarantine=enforce` + `min_release_age` | **403 held** | quarantine is first-seen on the *mirror* clock — an old artifact is "new to this mirror" and held for the TTL regardless of release age |
| `quarantine=observe` + `min_release_age` | **200 served** | min-age defers to quarantine even under `observe` → no freshness at all |

So there is **no** configuration in which a "do not serve artifacts released < N days ago" policy works. `trust_upstream_dates` (parsed for cargo/conan/go/npm/nuget/pub/pypi/terraform) has no effect, because min-age never activates. Verified live on PyPI (all three configs) and npm; the guard and the defer are registry-independent.

## Root cause

#742 conflated two distinct semantics:

- **release age** — how old the artifact is *upstream* (what `min_release_age` + `trust_upstream_dates` expressed);
- **first-seen cooldown** — how long since *this mirror* first saw the digest (quarantine, on NORA's own clock).

By subordinating min-age to quarantine *and* requiring quarantine for min-age, the release-age control became unreachable. A provably-old artifact is held as "new to this mirror."

## Proposed fix (pick one)

1. **(recommended) Quarantine records `first_seen` = the trusted upstream date** when `trust_upstream_dates=true` and the registry supplies one, instead of `now`. Then an old artifact is immediately `Mature` (served), a fresh one is held until `first_seen + TTL` — the two controls unify and `min_release_age` / `trust_upstream_dates` become meaningful again. Dateless registries (maven/gems/ansible/raw) fall back to the current first-seen behavior.
2. Make `min_release_age` an independent **active** control when `trust_upstream_dates` yields a real date (AND-compose with quarantine) instead of deferring.
[...]

Issue #748, opened 2026-06-19
Upstearm date not being used for PyPi packages


pyther:
### Description

sglang 0.5.11 was released on May 5, 2026
```
$ python3 -m pip download --index-url http://127.0.0.1:4000/simple/ --dest . sglang==0.5.11
Looking in indexes: http://127.0.0.1:4000/simple/
Collecting sglang==0.5.11
  ERROR: HTTP error 403 while getting http://127.0.0.1:4000/simple/sglang/sglang-0.5.11-cp313-cp313-manylinux_2_34_x86_64.whl (from http://127.0.0.1:4000/simple/sglang/)
ERROR: Could not install requirement sglang==0.5.11 from http://127.0.0.1:4000/simple/sglang/sglang-0.5.11-cp313-cp313-manylinux_2_34_x86_64.whl because of HTTP error 403 Client Error: Forbidden for url: http://127.0.0.1:4000/simple/sglang/sglang-0.5.11-cp313-cp313-manylinux_2_34_x86_64.whl for URL http://127.0.0.1:4000/simple/sglang/sglang-0.5.11-cp313-cp313-manylinux_2_34_x86_64.whl (from http://127.0.0.1:4000/simple/sglang/)
```

What Nora has in its quarantine file
```
$ cat /scratch/nora/storage/quarantine.jsonl
{"registry":"pypi","digest":"sha256:73a0130a0a0e1365f7dc5517648807b2da3cba3067259d993289e4458e3da5aa","first_seen":1781891960,"upstream":"https://files.pythonhosted.org/packages/b8/99/019d2d7a3b3dc7fe0c8de392473515855b07d2e652c31d68dfc5358232e6/sglang-0.5.13.post1-cp313-cp313-manylinux_2_34_x86_64.whl"}
{"registry":"pypi","digest":"sha256:db716eb7081082e2fb1e0542cd860a553243ea1bb1387925aa9d517d44e03b96","first_seen":1781891966,"upstream":"https://files.pythonhosted.org/packages/4c/03/d9fe8402d18232c3099e3b502903dd150b74ce199aed7348568b3d2fdffa/sglang-0.5.11-cp313-cp313-manylinux_2_34_x86_64.whl"}
```

Config
```
[server]
host = "127.0.0.1"
port = 4000
# public_url = "https://registry.example.com"  # Required when host = 0.0.0.0 or behind reverse proxy
body_limit_mb = 2048
trust_upstream_dates = true

[storage]
mode = "local"
[...]

PR #903, opened 2026-08-10
fix(npm): self-prime npm metadata on tarball download so `trust_upstream_dates` matures old npm artifacts

patsevanton:
## Problem

On NORA v1.1.0, `min_release_age` blocks **every** npm package as `0h old`
under `trust_upstream_dates=true` — even provably-old artifacts like
`lodash@4.17.21` (published 2021-02-17). The block surfaces through the
digest-quarantine path (`x-nora-quarantine: pending`, 14d TTL), making the npm
proxy unusable in `mode=enforce` and noisy in `mode=audit`. Cargo and PyPI are
unaffected: with the same flag they serve old releases immediately.

Refs: #902, #750, #748.

## Root cause

npm is the **only** dated registry without a self-prime on its download path:

| Registry | Date field            | Self-prime                          | Status   |
|----------|-----------------------|-------------------------------------|----------|
| PyPI     | `upload_time_iso_8601`| `ensure_pypi_dates_cached` (pypi.rs:793)           | present  |
| Cargo    | `created_at`          | `ensure_cargo_metadata_cached` (cargo_registry.rs:343) | present  |
| **npm**  | `time[version]`       | **none**                            | **absent** |

A locked/direct npm tarball fetch (`/npm/{name}/-/{name}-{version}.tgz`) can
arrive **without** a prior packument request, so `npm/{name}/metadata.json`
(the carrier of `time[version]`) is never cached. When `proxy_gate_dated`
(`digest_quarantine.rs:346`) runs, `extract_npm_publish_date` (`npm.rs:1171`)
returns `None` for `publish_ts`, `first_seen` is seeded from `now`, and the
artifact is held for the full quarantine TTL — regardless of its real upstream
release date.

Confirmed empirically in #902: after the `NORA_TRUST_UPSTREAM_DATES=true` env
workaround (helm chart `nora-0.4.4` does not render
`[server].trust_upstream_dates` in `config.toml`), Cargo and PyPI serve old
releases with HTTP 200, while npm still blocks with `x-nora-quarantine: pending`.

### New `ensure_npm_metadata_cached` (npm.rs)

Self-prime `npm/{name}/metadata.json` on the tarball download path, mirroring
`ensure_cargo_metadata_cached` (`cargo_registry.rs:343`) and
`ensure_pypi_dates_cached` (`pypi.rs:793`). Behavior:

- **No-op** if `metadata.json` is already cached.
- **Namespace-safe**: internal-namespace packages are never fetched upstream
  (#68 dependency-confusion guard).
- **Gated on proxy**: exits early if `npm.proxy` is unset.
- Fetches the packument from upstream via `proxy_fetch`, rewrites tarball URLs
  to point at NORA (via `rewrite_tarball_urls` / `replace_upstream_bytes`,
  exactly as `refetch_metadata` and the proxy path do), and writes it
  synchronously to storage so the date is readable on the **same** request.
- **Best-effort**: any failure (network, parse, storage) leaves the date
  `None` — fail-safe, the package is held as "new".

### Call site (npm.rs:315-326)

Before `extract_npm_publish_date`, when `trust_upstream_dates = true`, call
`ensure_npm_metadata_cached(&state, &package_name).await`. After this,
`proxy_gate_dated` receives `Some(publish_ts)` from `time[version]`,
`first_seen` is seeded from the real upstream publish date, and a
provably-old npm artifact matures immediately.

## Reproduction (before fix)

```bash
curl -D - -o /dev/null \
  -H "Authorization: Bearer nra_..." \
  https://nora.<host>/npm/lodash/-/lodash-4.17.21.tgz
```

Result: HTTP 403, `x-nora-quarantine: pending`, TTL 1209600s (14d).

## Expected behavior (after fix)

`lodash@4.17.21` (published 2021-02-17) is older than `min_release_age=7d`, so
`first_seen` is seeded from `time["4.17.21"]`, the artifact matures
immediately, and the request succeeds with HTTP 200.

## Scope

This PR fixes **npm only**. Cargo and PyPI already self-prime and are not
touched. Go (`.info.Time`) was not tested in #902 and is left for a follow-up
if needed.

## Out of scope

- Helm chart gap: `nora-0.4.4` does not render `[server].trust_upstream_dates`
  in `config.toml`. The `extraEnv: NORA_TRUST_UPSTREAM_DATES=true` workaround
  is still required until the chart is fixed. Tracked separately.
- Go registry self-prime (`.info.Time`).

## Verification

- [x] `cargo build -p nora-registry`
- [x] `cargo test -p nora-registry npm` — 108 passed, 0 failed
- [x] `cargo clippy -p nora-registry --lib` — no warnings
- [x] `cargo fmt -p nora-registry -- --check` — clean

## References

- Issue: https://github.com/getnora-io/nora/issues/902
- Diagnostic comment: https://github.com/getnora-io/nora/issues/902#issuecomment-5235444856
- `nora-registry/src/registry/npm.rs:1171` — `extract_npm_publish_date`
- `nora-registry/src/registry/npm.rs:315-326` — npm download-path date extraction (call site)
- `nora-registry/src/registry/cargo_registry.rs:343` — `ensure_cargo_metadata_cached` (reference)
- `nora-registry/src/registry/pypi.rs:793` — `ensure_pypi_dates_cached` (reference)
- `nora-registry/src/digest_quarantine.rs:346` — `proxy_gate_dated`
- `nora-registry/src/config/server.rs:27-58` — `trust_upstream_dates` default and docs
- CHANGELOG 0.9.5 (#748): Cargo self-primes `metadata.json`; npm was not mentioned.

### DIFF
--- a/nora-registry/src/registry/npm.rs
+++ b/nora-registry/src/registry/npm.rs
@@ -2101,6 +2101,342 @@ mod integration_tests {
         assert_eq!(json["username"], evil);
         assert!(json.get("admin").is_none());
     }
+
+    // ── ensure_npm_metadata_cached self-prime (#748) ──
+    //
+    // A tarball download (direct/locked fetch) arrives without a prior packument
+    // request, so metadata.json is uncached. ensure_npm_metadata_cached fetches
+    // the packument upstream and caches it (rewritten to NORA URLs) so the
+    // per-version `time[version]` date is readable on the same request.
+
+    /// Happy path: tarball cache-miss → ensure_npm_metadata_cached fetches and
+    /// caches the packument; tarball is also proxied and served. The cached
+    /// metadata must carry the upstream release date AND have its tarball URLs
+    /// rewritten to NORA.
+    #[tokio::test]
+    async fn test_npm_tarball_self_primes_metadata_on_cache_miss() {
+        use crate::test_helpers::{body_bytes, create_test_context_with_config, send};
+        use axum::http::{Method, StatusCode};
+        use wiremock::matchers::{method, path};
+        use wiremock::{Mock, MockServer, ResponseTemplate};
+
+        let upstream = MockServer::start().await;
+
+        // Packument (fetched by ensure_npm_metadata_cached → GET {proxy}/testpkg).
+        let packument = serde_json::json!({
+            "name": "testpkg",
+            "versions": {
+                "1.0.0": {
+                    "dist": {
+                        "tarball": format!("{}/testpkg/-/testpkg-1.0.0.tgz", upstream.uri())
+                    }
+                }
+            },
+            "time": { "1.0.0": "2020-01-15T10:30:00.000Z" }
+        });
+        Mock::given(method("GET"))
+            .and(path("/testpkg"))
+            .respond_with(
+                ResponseTemplate::new(200).set_body_string(packument.to_string()),
+            )
+            .mount(&upstream)
+            .await;
+
+        // Tarball (fetched by the cache-miss proxy path → GET {proxy}/testpkg/-/testpkg-1.0.0.tgz).
+        let tarball = b"TGZ-BYTES";
+        Mock::given(method("GET"))
+            .and(path("/testpkg/-/testpkg-1.0.0.tgz"))
+            .respond_with(ResponseTemplate::new(200).set_body_bytes(tarball.to_vec()))
+            .mount(&upstream)
+            .await;
+
+        let ctx = create_test_context_with_config(|cfg| {
+            cfg.npm.proxy = Some(upstream.uri());
+        });
+
+        // Direct/locked tarball fetch — no prior packument request.
+        let resp = send(
+            &ctx.app,
+            Method::GET,
+            "/npm/testpkg/-/testpkg-1.0.0.tgz",
+            "",
+        )
+        .await;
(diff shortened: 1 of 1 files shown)


devitway (approved):
LGTM — full gate pass.

**L0:** fmt ✓ clippy ✓ 1723 tests ✓ semver-checks ✓
**L1:** coherence-check.sh PASS (0 errors) · contract_gate 2 RED (pre-existing, unrelated)
**L2:** Regression map 12/12 PASS — config, mod, main, openapi, test_helpers, digest_quarantine, metrics, nora_base_url, all module boundaries clean
**L3:** Structural parity with `ensure_cargo_metadata_cached` / `ensure_pypi_dates_cached` confirmed — guard ordering, error handling, timeout, auth, CB, return type, call-site gate, concurrency model all match

**Findings (non-blocking):**
- No test for scoped packages (`@scope/name`) through the full self-prime→extract_date path (internal-namespace guard test exists but covers early-return, not the fetch path)
- `let _ = storage.put(...)` at line 1229 — silent on failure; a `tracing::warn!` would aid debugging (consistent with line 744)
- **Pre-existing gap found during parity audit:** `ensure_pypi_dates_cached` (pypi.rs:793) is missing the `is_internal_namespace` guard that Cargo and npm both have — filing as separate issue

ertime037 (approved):

PR #911, opened 2026-08-14
fix(docker): fail closed on transient storage errors in manifest reads

jgrund:
## Problem

The Docker manifest read path falls back from the namespaced storage key to the legacy (flat) key on **any** storage error, not just `NotFound` (`storage_get_with_fallback`, `storage_get_reader_with_fallback`, and the inline read in `get_manifest`).

When both keys hold copies — a proxied-current one under `docker/{ns}/…` and a pre-migration or hosted one under `docker/{name}/…` — a transient object-store error (S3/GCS 5xx, network blip) on the namespaced key silently swaps in the legacy copy **and** reclassifies the manifest as hosted, skipping tag revalidation. The same tag GET can answer with different bytes request-to-request, gated on backend weather. Downstream systems that pin the resolved digest see the tag flap.

A transient error on a hosted name's read could likewise fall through to the upstream proxy under the default-allow fallback, resolving a locally-owned name against an upstream where it may be squatted.

## Fix

Key fallback advances only on `StorageError::NotFound`; any other storage error fails closed with 500. Clients retry 5xx; they cannot un-see wrong bytes.

## Also

- `COMPAT.md`: new "Proxy cache and tag freshness" section documenting tag revalidation semantics (`docker.metadata_ttl`: -1/0 = revalidate every pull, positive = staleness window; digests immutable; `serve_stale` behavior; pushed tags are authoritative and never revalidated).
- `dist/nora.env.example`: document `NORA_DOCKER_METADATA_TTL`.

## Tests

- `manifest_get_fails_closed_on_transient_storage_error` — injected non-NotFound error on the namespaced key with a readable legacy copy must 500, not serve the legacy bytes (fails before this change).
- `manifest_get_serves_hosted_legacy_on_ns_not_found` — genuine NotFound still falls back to the hosted flat-key copy, served authoritatively with no upstream round trip and no `x-nora-stale`.

### DIFF
--- a/COMPAT.md
+++ b/COMPAT.md
@@ -27,6 +27,22 @@ This document describes which parts of each registry protocol are implemented in
 | Cross-repo blob mount | POST | — | Not implemented |
 | Referrers API | GET | — | OCI 1.1 referrers |
 
+### Proxy cache and tag freshness
+
+- Digest references (`sha256:…`) are content-addressed: cached forever, never
+  revalidated.
+- Tag references on **proxied** images are revalidated against the upstream on
+  every pull by default (`docker.metadata_ttl = -1`). A positive
+  `metadata_ttl` (seconds) is the staleness bound: within the window the
+  cached tag is served without an upstream round trip; after it, the next pull
+  revalidates. `0` or negative always revalidates.
+- If every configured upstream fails and `docker.serve_stale = true`
+  (default), the last cached manifest is served with `x-nora-stale: true`.
+- Locally **pushed** (hosted) images are authoritative: a pushed tag is served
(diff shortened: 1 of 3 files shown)


ertime037 (approved):

devitway (approved):
Reviewed against fail-closed invariant and proxy graceful-degrade. Fallback on NotFound-only is the correct fix — transient error serving a different copy is a data integrity violation. Tests with FailingPrefixBackend prove it. Note: after #915 merges, FailingPrefixBackend needs a copy() stub (integration catch from local merge test). LGTM.

PR #912, opened 2026-08-14
fix(auth): atomic token-file writes; store errors answer 503, not 401

jgrund:
## Symptom

Valid API tokens intermittently 401 (`Invalid username or password`) for ~seconds-to-a-minute, then recover — same creds, unchanged path. Only token-authenticated requests are affected (htpasswd loads at boot; OIDC is network-side), which makes it look like per-account auth flapping.

## Two defects

**1. Non-atomic token-file rewrites.** Three writers rewrite `{prefix}.json` in place with `fs::write` (truncate + write): the 30s `flush_last_used` background loop (every active token, every 30s), the SHA256→Argon2 migration on the verify path, and `create_token`. `verify_token`'s cache-miss path reads the same file concurrently — a reader between truncate and write sees a torn/empty file. Compounding: a failed verify never populates the verify cache, so every subsequent request re-reads disk until a clean read lands (a *burst* of 401s, not one), and `flush_last_used` skips unparseable files (`continue`), so a file left corrupt by an interrupted write is never repaired.

Fix: `write_token_file` — same-directory temp file, permissions set, `rename` over the target (atomic on POSIX). Listings skip the temp name (extension `tmp` ≠ `json`) and `is_valid_hash_prefix` keeps it out of revoke paths.

**2. Store errors read as credential verdicts.** Both the Bearer and Basic middleware paths collapse `TokenError::Storage` (I/O or parse failure) into the invalid-credentials 401 — and record it in the per-IP failure tracker, so a storage blip additionally walks well-behaved clients toward brute-force lockout. Both paths now answer **503** on `Storage` without recording a failure. Safe on the Bearer path: only `nra_`-prefixed tokens reach disk (`InvalidFormat` returns first), so a legitimate OIDC JWT can never be swallowed by the 503.

## Tests

- `verify_reports_storage_error_on_corrupt_file` — torn file ⇒ `Storage`, not `NotFound` (the classification the middleware depends on).
- `flush_last_used_replaces_file_atomically` — flush applies `last_used`, leaves a whole file, no temp leftover, token still verifies cold.
- `test_basic_auth_token_store_error_returns_503_not_401` / `test_bearer_token_store_error_returns_503_not_401` — full middleware paths via the test router.

Related: #911 (same fail-closed theme on the Docker manifest read path).

### DIFF
--- a/nora-registry/src/auth/mod.rs
+++ b/nora-registry/src/auth/mod.rs
@@ -398,6 +398,19 @@ pub async fn auth_middleware(
                     request.extensions_mut().insert(AuthenticatedRole(role));
                     return next.run(request).await;
                 }
+                // A store I/O/parse failure is not a credential verdict. Only an
+                // `nra_`-prefixed token reaches disk (anything else returns
+                // `InvalidFormat` first), so there is no OIDC identity to fall
+                // through to: answer 503 so the client retries, instead of a 401
+                // that makes valid creds flap during a storage blip.
+                Err(crate::tokens::TokenError::Storage(e)) => {
+                    tracing::error!(error = %e, "token store read failed during Bearer auth");
+                    return (
+                        StatusCode::SERVICE_UNAVAILABLE,
+                        "Token verification unavailable",
+                    )
+                        .into_response();
+                }
                 Err(_) => {
                     // Token verification failed — fall through to OIDC
                 }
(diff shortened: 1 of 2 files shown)


ertime037 (approved):

devitway (approved):
Reviewed against ADR-2/8, fail-closed invariant. Atomic token writes (tmp+rename) fix the torn-file race. Storage errors → 503 not 401 — correct fail-closed semantics. 4 tests cover both paths. LGTM.

PR #913, opened 2026-08-14
perf(rpm,deb): read index-rebuild sidecars concurrently

jgrund:
## Problem

`regenerate_repodata` (rpm) and `regenerate_indexes` (deb) run inside every publish request, and read one JSON sidecar per package from storage **sequentially**. Each PUT is therefore O(repo size) in storage round-trips. On an object-store backend this dominates: with ~1.7k packages at ~60ms per GET, every upload takes ~2 minutes server-side regardless of file size, and publishing a 40-package build spends over an hour in sidecar reads alone. Observed in production (GCS backend): uniform ~114s per uploaded RPM.

## Fix

Shared `read_json_sidecars` helper in `registry/mod.rs` — lists the sidecar prefix, fetches with `buffer_unordered(32)`, parses each. Completion order is unconstrained because both callers already sort records before generating indexes (the documented byte-identical determinism contract), so output is unchanged. ~1.7k reads drop from ~114s to a few seconds.

No behavior change beyond latency: same fail-closed contract (any read/parse error aborts the rebuild), same deterministic output. Existing rpm/deb suites cover the rebuild path (63 tests green).

### DIFF
--- a/nora-registry/src/registry/deb.rs
+++ b/nora-registry/src/registry/deb.rs
@@ -513,21 +513,7 @@ pub(crate) async fn regenerate_indexes(
     repo: &str,
 ) -> Result<(), String> {
     let meta_prefix = format!("deb/{repo}/{META_DIR}/");
-    let keys = storage
-        .list(&meta_prefix)
-        .await
-        .map_err(|e| format!("list sidecars: {e}"))?;
-
-    let mut pkgs = Vec::with_capacity(keys.len());
-    for key in &keys {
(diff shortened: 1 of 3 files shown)


ertime037 (approved):

devitway (approved):
Reviewed against ADR-4 (shared helper, not trait dispatch — fine). buffer_unordered(32) is reasonable. Callers still sort for deterministic output. LGTM.

PR #914, opened 2026-08-14
fix(cleanup): unify retention and GC into one scheduler so retention actually runs

jgrund:
## Problem

GC and retention ran as two independent tokio tasks with identical default intervals (24h), sharing one `cleanup_lock`. Both ticked at the same instant every cycle; the periodic path used `try_lock`, so GC took the lock and retention skipped with no retry:

```
19:40:00.838634  INFO nora::gc         GC scheduler: starting periodic run
19:40:00.838689  INFO nora::retention  Retention: cleanup lock held (GC or retention running), skipping
```

Same millisecond, identical on consecutive days. In a production deployment with both schedulers on the default 24h interval, age-based retention never ran.

## Fix

One periodic cleanup task replaces both schedulers (`nora-registry/src/cleanup.rs`):

- **One task, one lock hold per cycle.** Each cycle acquires `cleanup_lock` once and runs every due pass under it. The acquisition **waits** instead of `try_lock`-and-skip — the starvation race is gone by construction, since this task is now the only periodic contender. The lock is kept because the `cleanup_lock` → `publish_lock` ordering invariant and any future manual/admin cleanup entry point still need a guard. Both the sleep and the lock acquire race cancellation, so SIGTERM stays prompt.
- **Retention → GC order.** Retention deletes expired versions and files, creating exactly the orphans the GC pass then sweeps, so one cycle reclaims what the split design needed two cycles for.
- **Per-pass intervals honored.** Each pass keeps its own configured interval; the next wake is the earliest due time. Due times are computed from the wake instant, so same-interval passes stay phase-locked and chain in every cycle, and a pass overrunning its interval fires again on the next loop.
- **Boot cycle preserved.** Every pass is due at boot — a process restarting more often than its interval must still clean.
- **Panic isolation.** A panicking pass is caught: it can neither skip the later passes of its own cycle nor kill the scheduler.

Per-pass logging and the retention audit entry are unchanged. `spawn_gc_scheduler` and `spawn_retention_scheduler` are deleted along with their scheduler tests; `run_gc` and `run_retention` are untouched.

## Tests

- `cycle_runs_passes_in_order_at_boot` — the first cycle fires at boot (well under the interval) and records retention before GC.
- `pass_panic_does_not_prevent_later_passes` — a pass that panics every cycle does not stop the later pass in the same cycle, and the scheduler survives into the next cycle.
- `cycle_waits_for_cleanup_lock_instead_of_skipping` — with the lock held nothing runs; on release both passes run, so the periodic path parks instead of skipping.
- `cancel_while_parked_on_lock_stops_scheduler` — cancellation while parked on the lock stops the task promptly, with no pass run.

### DIFF
--- a/nora-registry/src/cleanup.rs
+++ b/nora-registry/src/cleanup.rs
@@ -1,0 +1,249 @@
+//! Unified periodic cleanup scheduler.
+//!
+//! Retention and GC used to run as two independent tasks with the same
+//! default interval, sharing one `cleanup_lock`. Both ticked at the same
+//! instant every cycle and the periodic path used `try_lock`: GC won the
+//! race every time and retention logged "cleanup lock held ... skipping"
+//! with no retry, so age-based retention never ran on a default schedule.
+//!
+//! One task removes that race by construction — every due pass runs in
+//! order under a single lock acquisition, and the acquisition waits instead
+//! of skipping. Retention runs before GC because retention deletes expired
+//! versions, which creates the orphans GC then sweeps: one cycle now
+//! reclaims what the split design needed two cycles for.
+
+use futures::future::BoxFuture;
+use futures::FutureExt;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::Instant;
+use tokio_util::sync::CancellationToken;
+use tracing::warn;
+
+pub(crate) struct CleanupPass {
+    pub name: &'static str,
+    pub interval: Duration,
+    /// Runs one pass. Owns its own start/done logging.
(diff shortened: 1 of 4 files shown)


ertime037 (approved):

devitway (approved):
Reviewed against ADR-1/4. Race eliminated by construction — one scheduler, ordered passes, no try_lock skip. catch_unwind per pass, CANCEL-SAFETY on every select!. Net -150 lines. Retention->GC order is correct (expired versions create orphans GC sweeps). LGTM.

PR #915, opened 2026-08-16
feat(docker): cross-repo blob mount (?mount=&from=)

jgrund:
## Problem

`POST /v2/{name}/blobs/uploads/?mount=<digest>&from=<repo>` was ignored: the query parameters were dropped and every mount attempt opened a normal upload session, so promoting an image between repositories on the same instance re-uploaded every layer even though the bytes were already in the bucket.

## Fix

Cross-repo blob mounting per the OCI distribution spec:

- **`StorageBackend::copy(src, dst)`** — server-side copy inside the backend. Object stores delegate to `ObjectStore::copy` (S3 `CopyObject` / GCS rewrite — bytes never transit the process); local storage hard-links, falling back to a file copy across devices, and fsyncs the parent dir per the durability invariant.
- **`Storage::copy(src, dst, sha256)`** pins the destination key from the known digest via `record_hash` (no read-back), inheriting the source pin when no digest is supplied, and fails closed on a pin-record failure exactly like `put`/`put_from_path` — a copied-but-unpinned blob would otherwise be served unverified.
- **`mount_blob`** handler: dispatch recognizes `mount`+`from` on the upload POST. A mountable blob answers `201 Created` with `Location` and `Docker-Content-Digest` (response shape shared with `upload_blob` via `blob_created`). Every reason a mount cannot happen — unknown digest, unusable source repo, storage failure, a digest still inside its proxy-quarantine cooldown — degrades to the normal `202` upload session the spec requires, never a new 4xx. Namespace write scope on the target is enforced by the existing dispatch gate; reads stay ungated, matching blob GET.
- OpenAPI: `mount`/`from` params and the `201` response documented on the upload POST.

## Tests

- `test_docker_cross_repo_blob_mount` — push to `srcrepo`, mount into `dstrepo`: 201 with correct headers, blob readable from `dstrepo` byte-for-byte; re-mount with an empty source still 201 off the destination copy.
- `test_docker_mount_unknown_blob_falls_back_to_upload` — unknown digest degrades to 202 with an upload session.
- `test_docker_mount_invalid_source_repo_falls_back_to_upload` — invalid source repo degrades to 202, not 4xx.
- `copy_duplicates_bytes_and_pins_destination`, `copy_missing_source_is_not_found` — storage-level copy semantics and pin propagation.

### DIFF
--- a/nora-registry/src/gc.rs
+++ b/nora-registry/src/gc.rs
@@ -1110,6 +1110,9 @@ mod tests {
         )> {
             Err(crate::storage::StorageError::NotFound)
         }
+        async fn copy(&self, _src: &str, _dst: &str) -> crate::storage::Result<()> {
+            Err(crate::storage::StorageError::NotFound)
+        }
     }
 
     /// #610 (hardening for #584): an orphan whose age cannot be determined
--- a/nora-registry/src/openapi.rs
+++ b/nora-registry/src/openapi.rs
@@ -455,14 +455,19 @@ pub async fn docker_manifest_delete() {}
 /// Start blob upload
 ///
 /// Initiates a resumable blob upload. Returns a Location header with the upload URL.
+/// With `mount` + `from`, links a blob the source repository already holds instead;
+/// if it cannot be mounted, an upload session is started as usual.
 #[utoipa::path(
     post,
     path = "/v2/{name}/blobs/uploads/",
     tag = "docker",
     params(
-        ("name" = String, Path, description = "Repository name")
+        ("name" = String, Path, description = "Repository name"),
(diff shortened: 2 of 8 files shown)


ertime037 (dismissed):

devitway (dismissed):
Reviewed against ADR-2/3/8, StorageBackend trait extension, fail-closed invariant, module boundaries. Local copy uses hard_link (zero extra bytes, correct). S3 uses CopyObject (server-side). Every mount failure degrades to upload (202) per OCI spec. Quarantine gate prevents digest laundering. Pin-store pins dst fail-closed. Minor: S3 CopyObject has 5GB limit — worth documenting as known limitation for very large blobs. LGTM.

devitway:
Reopening from main repo branch to fix verified signatures + code owner review.

PR #947, opened 2026-08-31
feat(storage): hash pins on S3/GCS via object metadata

jgrund:
## What

Hash pins on S3/GCS. The SHA-256 integrity pin was a local-filesystem-only feature (`Storage::new_s3` / `new_gcs` set `pin_store: None`): object-store deployments had no at-rest integrity verification, and raw files had no `ETag`, so `If-None-Match` (304) and `If-Match "<etag>"` conditional overwrite could not work there.

Pins are now a backend concern, stored with the bytes:

- **Local**: `LocalStorage` owns the `HashPinStore` (same `.nora-pins.ndjson` path and format — no migration).
- **S3/GCS**: the pin is the user-defined `sha256` object metadata (`x-amz-meta-sha256` / `x-goog-meta-sha256`), written atomically with the object via `put_opts` / `put_multipart_opts` and read back from `GetResult.attributes` on GET/HEAD. Server-side `copy` carries user metadata, so a copied blob inherits its pin.
- **`Storage` wrapper**: validation, metrics, and the fail-closed verify gate only. `StorageBackend::get` / `get_reader` return `(bytes, pin)` from one round-trip, so the bytes and the pin are always the same object version (no HEAD-then-GET skew across a raw overwrite). `get_pin_hash` → `async fn pin`.
- **raw**: `ETag` / `If-None-Match` / `If-Match` now work on every backend; the GET path opens the reader first and answers 304 from the returned pin.
- `re-pin` works on every backend; on an object store it rewrites the object (metadata cannot be changed in place).

Objects written before this change carry no metadata and stay open-world (`GateOutcome::Unpinned`), exactly as today, until rewritten.

Removed as dead: `RepinOutcome::NoPinStore`, `Storage::pinned_count`, `HashPinStore::{record,verify,len}`, the `pin_error` metric label, the two "Hash pin store disabled" startup warnings, and the S3 no-pin caveats in `import` and `verified.rs` docs.

## Why

Integrity verification and raw `ETag` should not depend on the storage backend; object stores already have a natural, atomic place for the pin. Verified against `object_store` 0.13.2: metadata attributes are sent on put and multipart-init and parsed on GET/HEAD for both the AWS and GCP clients and for `InMemory`.

## Testing

- New object-backend tests (`ObjectStorage::in_memory()`): put/get pin round-trip, `put_from_path` with/without digest, `copy` inherits, tampered bytes under a valid pin → `IntegrityViolation`, object without metadata → open-world, `repin` mismatch/update.
- New raw integration test over the object backend: HEAD `ETag` == sha256, `If-None-Match` → 304, `If-Match` stale → 412, `If-Match` match → 200 with the new ETag.
- `tests/s3-backends/test.sh` gains the same HEAD-ETag and 304 checks for MinIO / SeaweedFS / Garage.

## Checklist

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -99,6 +99,13 @@ matched the bytes, or `Unpinned` for an open-world key with no pin. Callers must
 verified one. A pin mismatch returns `IntegrityViolation` rather than the bytes,
 so the gate fails closed.
 
+The pin itself belongs to the backend, which keeps it with the bytes: an NDJSON
+sidecar (`.nora-pins.ndjson`) on the local filesystem, and the user-defined
+`sha256` object metadata on S3/GCS, written atomically with the object and read
+back on GET/HEAD. The wrapper only validates keys and runs the gate; an object
+stored without a digest (or written before pins existed) simply has none and
+stays open-world.
+
 The curation layer is a second trust boundary for proxy traffic. When mode is
 `enforce`, a package must pass all filters (blocklist, allowlist, namespace,
 integrity) before reaching storage. When mode is `audit`, blocked packages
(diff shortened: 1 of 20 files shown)


jgrund:
The failing `Check Markdown Links` job is unrelated to this change: GitHub now 404s every repository's `/stargazers` page for logged-out clients, so the stars badge in README breaks the checker on all PRs. Fix in #948; this check should pass on a re-run once that merges.

PR #951, opened 2026-09-03
feat(storage): hash pins on S3/GCS via object metadata

devitway:
Relands #947 (by @jgrund), rebased onto current `main`.

Records the SHA-256 integrity pin as object metadata (`x-amz-meta-sha256` / `x-goog-meta-sha256`) on the S3 and GCS backends. The pin is written atomically with the object (on `put_opts` / the multipart init), so an object is never stored without its pin; `get` re-verifies the pin on read and fails closed on a mismatch. Raw uploads are additionally validated against an RFC 9530 `Repr-Digest` header when present.

Also adds a direct `validate_storage_key()` at the top of `Storage::get()` for defense in depth.

Closes #947

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -99,6 +99,13 @@ matched the bytes, or `Unpinned` for an open-world key with no pin. Callers must
 verified one. A pin mismatch returns `IntegrityViolation` rather than the bytes,
 so the gate fails closed.
 
+The pin itself belongs to the backend, which keeps it with the bytes: an NDJSON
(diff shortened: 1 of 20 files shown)


ertime037 (approved):

devitway:
Superseded by #952 — same reviewed diff re-landed as one web-flow-signed commit (main requires verified signatures; these commits were unsigned). Closing.

Note: the `auto-reland` workflow failed with `Bad credentials (401)` — the `RELAND_TOKEN` secret is expired and needs renewing for the automated path.

Issue #956, opened 2026-09-05
npm hosted: packument is not rebuilt from versions/ — 404 while every published version sits in storage


devitway:
Found while validating #889 on a polygon. Credit to @ivaseeq — his packument work surfaced this; the defect is ours, on `main`.

> **Corrected 2026-09-05.** The first version of this issue claimed a 31-second storage scan before the 404. That was a mismeasurement: the 31 s was a one-off 30 s upstream timeout to `registry.npmjs.org` on the first request after process start. With the proxy disabled the same request answers **404 in 0.98 ms**. There is no fan-out here. The correctness defect below is real and unchanged; the performance claim is withdrawn.

## Problem

When a hosted npm package has its per-version objects in storage but no materialized `npm/<pkg>/metadata.json`, `GET /npm/<pkg>` answers **404** — while every published version is still sitting in the bucket.

The packument is derived state, and the code to derive it already exists: `regenerate_packument()` in `registry/npm.rs` lists `npm/<pkg>/versions/`, reassembles dist-tags and package fields, and writes `metadata.json`. Today it is only ever called from the publish path. The read path treats a missing `metadata.json` as "no such package" and falls through to the upstream proxy, which then 404s for a package that only exists locally.

`ARCHITECTURE.md` states the invariant this violates: derived indexes are rebuildable from what is on disk. Here the rebuild exists and simply is not reached.

Reachable in practice through direct-storage import and migration, which is where it was originally reported.

## Reproduction

NORA `0644178`, S3 backend (MinIO), `auth.enabled=false`.

1. `npm publish` 500 versions of a package into the hosted npm registry.
2. Confirm the layout: `npm/<pkg>/metadata.json`, `npm/<pkg>/versions/*.json` (500 objects), tarballs — 1503 objects.
3. Delete `npm/<pkg>/metadata.json`. Leave `versions/` untouched.
4. Restart NORA and `GET /npm/<pkg>`.

## Observed

```
NORA_NPM_PROXY=""                       404, 0 B, 0.98 ms
NORA_NPM_PROXY=registry.npmjs.org       404, 0 B, 0.29 s   (the extra time is the upstream miss)
```

500 version objects remain in `npm/<pkg>/versions/`. The registry reports the package does not exist.

## Expected

[...]

PR #976, opened 2026-09-08
fix(npm): do not cache a proxied packument over a locally owned key

### LINKED ISSUE #975: bug(npm): proxied packument can overwrite a locally assembled one after the publish_lock is released
Surfaced by #974: once `lock-audit`'s false positive is removed, Check 2 reaches a
different write in `npm_handler` and reports it.

```
FAIL npm.rs:578 _guard drops (conditional/loop scope) before storage write at line 688
```

## The path

On a metadata request `key` is `npm/{name}/metadata.json` (`npm.rs:294`). The
packument-rebuild block added in #956 takes `publish_lock(&key)` at `npm.rs:577-578`
and releases it when `if has_versions` closes at `npm.rs:607`. It falls through
without returning when

- `regenerate_packument` returns `Err(())` — its only failing step is the
  `storage.put` of `metadata.json` (`npm.rs:1192-1196`), i.e. a storage write error; or
- the rebuild succeeds but the immediate read-back `storage.get(&key).await.ok()`
  (`npm.rs:583`) yields `None` — a read-after-write miss or a transient backend error.

Execution then continues into the proxy path, which fetches the upstream packument,
rewrites tarball URLs (`npm.rs:664-680`) and caches the result:

```rust
// npm.rs:683-693
let key_clone = key.clone();
tokio::spawn(async move {
    if let Err(e) = storage.put(&key_clone, &data_to_cache).await { … }
});
```

That writes the same `npm/{name}/metadata.json`, from a detached task, with the lock
already released.

## Why it matters

Two consequences, and the second needs no concurrency at all.

**A locally published package can lose its versions from the packument.** The block is
only entered when `versions/` is non-empty — the package *has* local publishes. On
fall-through the upstream copy is written over the local packument key, so a subsequent
`GET` serves a document that lists only upstream versions. It does not self-heal: the
read-path rebuild triggers on `metadata.json` being *absent*, and it now exists with
the wrong content. `npm install pkg@<locally-published-version>` fails until the next
publish or a retention-triggered regeneration rewrites the packument.

**It can also clobber a concurrent publish.** Publish assembles the packument under the
same `publish_lock`; the detached write above is unserialized against it and can land
afterwards, replacing a freshly assembled packument with the upstream one.

The trigger needs a storage write error or a read-after-write miss, so the likelihood
is low — but the outcome is silently wrong metadata that persists.

## Possible directions

- On the metadata path, do not cache the proxied packument at the local packument key
  when a local rebuild was attempted (`has_versions == true`) — the two documents
  describe different things and the local one is authoritative for that key.
- Or perform the cache write under the same `publish_lock`, re-checking the key after
  acquiring it, so it cannot overwrite a newer locally assembled packument.
- Independently: consider making the read-path rebuild trigger when the packument is
  missing versions that exist under `versions/`, not only when it is absent, so a
  packument that has been overwritten recovers on the next read.

`lock-audit` reports this today (after #974). If the write turns out to be safe for a
reason worth writing down, the alternative is a `// LOCK-SAFE: <reason>` marker plus
support for it in the check — happy to send either change.

devitway:
Closes #975. Surfaced by #974, once `lock-audit`'s false positive stopped hiding it.

## The defect

`npm/{name}/metadata.json` is the assembled form of keys this registry owns, but the
proxy branch cached the upstream document there unconditionally. On the metadata path
that branch is reached **with local versions present** only by falling through the
rebuild added in #956 — its write failed (`npm.rs:1192-1196` is the only failing step of
`regenerate_packument`), or the read-back right after it missed. The cached upstream
document then drops every locally published version from the packument.

It does not self-heal: the read-path rebuild triggers on the key being *absent*, and
after such a write it exists with the wrong content, so
`npm install pkg@<locally-published-version>` keeps failing until the next publish or a
retention-triggered regeneration. The write also lands unserialized against the
`publish_lock` released when the rebuild block closes, so a concurrent publish that
assembled the packument under that lock can be overwritten.

## The change

The proxy branch skips the cache write when the metadata path has local versions. The
upstream bytes are still served for that request; only the write is skipped, so the next
request retries the rebuild instead of finding a wrong document cached.

The write also carries a `// LOCK-SAFE:` reason now, because in every case that reaches
it the guard was never taken: either `is_tarball` (the `if !is_tarball` block was never
entered) or `!has_local_versions` (the `if has_versions` block was never entered).

## Verification

**Unit** — `npm_proxy_does_not_cache_upstream_packument_over_local_versions` injects the
fall-through with a backend whose reads of the packument key always miss, and records
every write to that key. The assembled document must be there (the positive control that
the rebuild ran) and the upstream one must not. On the unfixed code both writes land, the
upstream one second:

```
found: [("npm/hybrid/metadata.json", {"dist-tags":{"latest":"1.0.0"},"versions":{"1.0.0":…}}),
        ("npm/hybrid/metadata.json", {"description":"upstream-only-packument","versions":{"9.9.9":…}})]
```

**End to end** — MinIO behind a proxy that returns 404 for reads of
`npm/poly975/metadata.json` (the same fall-through, injected at the S3 layer), a real
`npm publish` of 1.0.0, then one packument request:

| | after publish | after the request |
|---|---|---|
| `main` | versions `[1.0.0]` | versions `[9.9.9]`, upstream marker present |
| this PR | versions `[1.0.0]` | versions `[1.0.0]`, no marker |

Both serve `200`; only the write differs.


ertime037 (dismissed):

devitway:
Rebased onto `main` after #974 landed — the conflict was the shared `## [Unreleased]` section in CHANGELOG.md, and both entries are now kept under one `### Fixed` heading.

Worth noting for the reviewer: with #974 already on `main`, this branch is the first tree where `make check` is green again — `lock-audit` reports nothing, because the check now understands branch exclusivity and the remaining write carries a `// LOCK-SAFE:` reason that is only valid once this fix makes it true.

```
$ make check
…
=== All checks passed ===
```

ertime037 (approved):

Issue #969, opened 2026-09-07
1.3.0: Very slow /simple/<package-name>/ requests


9128305:
### Description

With nora 1.3.0, uv is constantly failing with timeout errors when accessing `/simple/<package-name>/`. This did not happen before upgrading to 1.3.0.

### Expected Behavior

No timeout errors

### Steps to Reproduce

Run any uv command, for example:
```
uv lock
```

### NORA Version

1.3.0

### Registry Protocol

PyPI

### Storage Backend

S3-compatible

### Relevant Logs

```shell

```

devitway:
Thanks for the report — the cause is identified and a fix is open in #970.

In 1.3.0 the PEP 700 fields added to the simple JSON made `GET /simple/{name}/`
issue one storage `stat()` per file in the index, serially. On an S3 backend that
is a HEAD round-trip per file, so the response time scales with the number of files
the index lists. uv always requests the JSON index, which is why this is uv-only —
the HTML index pip uses was never on that path.

With the fix, a 40 000-file proxied index on S3 goes from 129.8 s and 59 764 storage
HEAD requests (failing after uv's retries) to 0.5 s and 2, matching 1.2.0.

To be sure your case is fully covered, could you share:

1. A package where it reproduces — or roughly how many files its index lists?
2. What the PyPI upstream is: pypi.org directly, or another index/proxy in front of it?
3. Roughly how many files are already cached locally for that package?
4. Is the S3 endpoint on the same network as NORA, or remote (rough round-trip latency)?

Question 3 matters because two costs on this path predate 1.3.0 and this fix does not
remove them: one `.sha256` GET per locally cached file of the package, and an upstream
index fetch on every request. Neither scales with the size of the upstream index, so
neither is the regression you hit — but if a package has a few thousand files cached
locally, some latency will remain after the fix, and it would be worth tracking
separately.

If you are able to build the branch (`fix/pypi-simple-json-nplus1-969`), a before/after
on your own setup would be the most direct confirmation.

PR #978, opened 2026-09-08
perf(npm): rebuild the packument with a concurrent read, not one key at a time

devitway:
`regenerate_packument` walked `versions/` and `dist-tags/` one key at a time. On an
object store every key is a network round-trip, and #956 moved that walk onto a read
path — a `GET /npm/{name}` whose `metadata.json` is missing rebuilds it — so the first
request after a packument goes missing pays one round-trip per version.

Measured against a loopback S3 store (MinIO), 500 published versions:

| request | time | S3 round-trips |
|---|---:|---|
| warm (`metadata.json` present) | 0.013 s | 1 GET + 1 HEAD |
| cold (`metadata.json` deleted) | 0.77 s | **505 GET** + 1 PUT |
| next request | 0.003 s | 1 GET |

The cost is `N+5` reads, once per package, and it self-heals — the rebuild writes the
packument, so a client that timed out finds it on the next try. It is linear in versions
and a remote store multiplies it by the round-trip time, which is what makes it worth
fixing even though it is not a regression.

## Change

The reads go through the same buffered fan-out the RPM and Deb index rebuilds use
(#913). Those call `read_json_sidecars`, which drops the key and fails the batch on the
first bad entry — a packument needs neither. The key *is* the version, and one
unreadable version must not turn a partial packument into no packument, which is exactly
what the sequential loop already did by skipping it. `read_keyed_blobs` is that variant:
same concurrency, keeps the key, skips what it cannot read.

The assembled document is unchanged: `serde_json::Map` is a `BTreeMap` here
(`preserve_order` is not enabled), so completion order never reaches the output.

## Test

A backend that records how many reads are in flight at once. A sequential walk never
exceeds one; the test asserts the peak is above that and that all eight versions still
land in the assembled document. Against the previous code it fails with
`peak in-flight reads was 1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 ## [Unreleased]
 
 ### Changed
+- **npm rebuilds a packument concurrently (#956 follow-up)** — `regenerate_packument` walked `versions/` and `dist-tags/` one key at a time. On an object store each key is a round-trip, and #956 put that walk on a read path, so the first request after a packument went missing paid one round-trip per version: 505 reads and 0.77 s for a 500-version package against a loopback S3 store, linear from there. It now reads them through the same buffered fan-out the RPM and Deb index rebuilds use, keeping the lenient behaviour — an unreadable or unparsable version is skipped rather than failing the rebuild.
+
+### Changed
 - **Every storage round-trip is counted (#969 follow-up)** — `stat`, `pin`, `list` and `list_with_meta` reached the backend without touching `nora_storage_operations_total`, which is why a handler issuing one `stat()` per file (tens of thousands of HEAD requests on a single PyPI index response) moved no metric and could only be found by reading code. All four now increment it; on `stat`/`pin` an absent object or an unpinned one is `status="miss"`, so ordinary misses do not inflate error-rate alerting. Counters only — no behavioural change.
 
 ### Fixed
--- a/nora-registry/src/registry/mod.rs
+++ b/nora-registry/src/registry/mod.rs
@@ -95,6 +95,34 @@ pub(crate) async fn read_json_sidecars<T: serde::de::DeserializeOwned>(
         .await
 }
 
+/// Read every key under `prefix` concurrently, returning the pairs that came back.
+///
+/// [`read_json_sidecars`] drops the key and fails the whole batch on the first bad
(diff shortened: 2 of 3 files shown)


ertime037 (approved):

devitway:
Holding this one — the end-to-end measurement does not reproduce the win yet, and the PR body claims one, so treat the numbers there as unverified until I replace them.

Against a loopback MinIO with a counting proxy in the path, 500 published versions, cold rebuild:

| | time | S3 reads |
|---|---:|---|
| `main` | 0.789 s | 505 |
| this branch | 1.473 s | 505 |

The read count is expected to be identical — concurrency overlaps round-trips, it does not remove them — but the wall clock went the wrong way. The unit test does show the fan-out happening (peak in-flight reads above one), so the suspect is the measurement rig: the counting proxy is a thread-per-connection Python server with `Connection: close`, which is fine under a sequential walk and a bottleneck under a 32-way one. Re-measuring with NORA pointed straight at MinIO.

If the win does not hold without the proxy either, this PR is not worth taking on performance grounds and I will say so.

devitway:
Corrected measurement, replacing the numbers in the PR body and the caveat in my previous comment.

The earlier run put a counting HTTP proxy between NORA and MinIO so it could report round-trips. That proxy is a thread-per-connection Python server with `Connection: close` — invisible under a sequential walk, and the bottleneck under a 32-way one. It inverted the result. Measured with NORA pointed straight at MinIO, 500 published versions, cold rebuild, three runs each:

| run | `main` | this change |
|---|---:|---:|
| 1 | 0.189 s | 0.052 s |
| 2 | 0.181 s | 0.049 s |
| 3 | 0.181 s | 0.050 s |
| mean | **0.184 s** | **0.050 s** |

3.7× on a loopback store, where per-read latency is near zero and the walk is CPU- and syscall-bound. On a remote store the gain scales with round-trip time toward the concurrency factor, since the cost there is latency rather than work.

The read count is unchanged at 505 either way — concurrency overlaps round-trips, it does not remove them. Apologies for the noise: the rig was built for #969, where the metric was the *count* of round-trips and a sequential walk never loaded it, and I reused it for a measurement about *overlapping* them without checking that it does not serialize first.

Issue #595, opened 2026-06-02
Proxy: coalesce concurrent upstream fetches on cache miss (single-flight)


devitway:
## Problem

On a cache miss — either a first-time proxy request or right after `cache_ttl::is_within_ttl()`
(`cache_ttl.rs:19`) returns `false` — every concurrent client that asks for the same artifact
independently goes to the upstream and independently writes the result to storage. There is no
coordination between in-flight fetches for the same key.

This is a classic thundering-herd / cache-stampede. It bites hardest exactly where it hurts:

- A popular metadata index (npm packument, PyPI `/simple/<pkg>/`, crates sparse-index line,
  Maven `maven-metadata.xml`) expires, and a CI fleet that pulls on every build sends N
  simultaneous requests → N upstream round-trips for one object.
- Each of those N responses then races to write the same storage key. Our writes are atomic
  (temp + rename), so we don't corrupt anything, but we do N redundant downloads and N redundant
  storage writes for a single logical refresh, multiplying upstream load and disk churn.

We already have the building blocks of in-flight coordination elsewhere, just not on the live
proxy path:

- `circuit_breaker.rs:42` tracks a single `half_open_in_flight` probe — proof we already reason
  about "only one request should go out right now."
- `mirror/npm.rs:187` deduplicates metadata fetches, but only inside the offline `nora mirror`
  CLI, by target, not for live concurrent requests.

The hot serving path (`registry/mod.rs::proxy_fetch`, `:173`) has none of this.

## Proposed solution

Add a small, in-memory **single-flight** coordinator for the proxy-miss path: for a given fetch
key, exactly one task ("leader") performs the upstream fetch + store; concurrent tasks
("followers") wait for it and then read the now-populated cache instead of fetching themselves.

Sketch:

- New `src/proxy_coalesce.rs` exposing:
  ```rust
  // Key = the storage key (or upstream URL) being fetched.
  pub struct InflightMap { /* Mutex<HashMap<String, Arc<Notify>>> */ }

  /// Run `fetch` under single-flight. Leader executes it; followers await and return
  /// `Coalesced::Waited` so the caller re-reads from storage.
  pub async fn coalesced<F, Fut, T>(map: &InflightMap, key: &str, wait_budget: Duration, fetch: F)
      -> Coalesced<T>
  where F: FnOnce() -> Fut, Fut: Future<Output = T>;
  ```
- RAII `FetchGuard` that removes the key from the map and `notify_waiters()` **on drop** — so a
[...]

PR #1002, opened 2026-09-16
fix(docker): stream a proxied blob while it is still being spooled

### LINKED ISSUE #1001: Docker proxy: a cache miss sends no response until the whole blob is spooled, so containerd ≥ 2.3 (30 s ResponseHeaderTimeout) never caches large layers
## Summary

On a Docker blob cache miss, `download_blob` fetches the whole blob from upstream into a temp file, uploads it to storage, and only then builds the HTTP response. No byte, not even the status line, reaches the client until the blob is completely cached.

containerd 2.3 added a hard-coded `ResponseHeaderTimeout: 30 * time.Second` to its registry transport (`core/remotes/docker/registry.go`, `DefaultHTTPTransport`; 2.1 has none). It is not overridable from `hosts.toml`, which only exposes `dial_timeout`, and it is a different timer from `image_pull_progress_timeout`. Any blob whose upstream fetch takes longer than 30 s therefore never reaches a containerd 2.3 node through NORA:

1. containerd requests the blob, NORA starts spooling, sends nothing.
2. At 30 s containerd's transport times out waiting for headers and retries once (another 30 s), then the pull fails and kubelet backs off.
3. Each abandoned request drops the handler future, so `TempFileGuard` deletes the partial spool. Nothing is cached.
4. The next attempt starts from zero. Repeat forever.

## What we observed

Alibaba Cloud ACK cluster, Kubernetes 1.36.2 (containerd 2.3.4), NORA 1.0.1 in `NORA_STORAGE_MODE=s3` (OSS), upstream `asia-docker.pkg.dev` reached over a ~100–150 Mbit/s single-stream cross-border link.

- One 1.27 GB image layer: **56 `Proxy blob download started` lines over four days, zero `Proxy blob download complete`**. Every deploy that needs it fails its rollout deadline. Two other >1 GB layers show the same loop.
- The bucket holds no blob larger than ~400 MB written after the nodes moved to containerd 2.3.4; smaller blobs, which finish inside 30 s, cache fine.
- A sibling cluster on containerd 2.1.9 (no header timeout) against the same NORA build caches new 1.27 GB layers daily.
- One failed rollout moved 17.6 GB across the link for an image whose real delta was 76 MB, because every retry of every uncached layer re-fetched from scratch.

## Proposal

Answer as soon as upstream has answered, and finish the fill regardless of the client:

- Split `fetch_blob_from_upstream` into `open_blob_from_upstream` (request, token auth, status check — returns once the upstream headers are in) and `spool_upstream_blob` (body → temp file, per-chunk `read_timeout`, incremental SHA-256). The existing function stays as the composition of both for the mirror path.
- In `download_blob`, spawn the spool + `put_from_path` as a detached task and return `200` immediately with `Content-Length` from upstream (when known) and `docker-content-digest`. The body follows the spool file as it grows (a second read handle on the temp file, driven by a `watch` channel of bytes-flushed / done), wrapped in the existing `VerifyingReader` so the streaming serve path keeps its digest check at EOF.
- A client that disconnects mid-fill no longer aborts the fill: the spool completes and the blob is cached, so containerd's retry is a cache hit instead of another cold start.

### Behaviour change to discuss

Today a digest mismatch from upstream is a `502` with no bytes sent ("verify-before-serve, no tee to client", #580). With early headers the digest is only known at the end, so a mismatch becomes a `200` whose body aborts before a clean end, and the bytes are still never cached. Registry clients verify the digest of every blob themselves, and the `VerifyingReader` already errors the response at EOF on the local-hit path, so this is the same failure mode NORA accepts there. If holding the body until verification is preferred, note that it does not help large blobs: containerd's 5-minute pull-progress timeout then cancels anything that spools slower than that.

I have a PR implementing this with tests for headers-before-body, fill-survives-disconnect, and the updated mismatch case.

synchrone:
Closes #1001.

## Problem

On a Docker blob cache miss, `download_blob` downloaded the whole blob to a temp file, stored it, and only then built the response. The client received nothing, not even the status line, until the blob was fully cached.

containerd 2.3 hard-codes `ResponseHeaderTimeout: 30s` on its registry transport (`core/remotes/docker/registry.go`; 2.1 has none, `hosts.toml` cannot override it, and `image_pull_progress_timeout` is a different timer). Any blob that takes longer than 30 s to fetch from upstream therefore failed on every attempt, and because each abandoned request dropped the handler future, `TempFileGuard` deleted the partial spool and the next attempt started from zero. Details and production numbers in #1001: a 1.27 GB layer was started 56 times in four days and never completed once, while a sibling cluster on containerd 2.1 cached the same layers daily.

## Change

- `fetch_blob_from_upstream` is split into `open_blob_from_upstream` (request, token auth, status check; returns once the upstream headers are in) and `spool_upstream_blob` (body → temp file with the per-chunk `read_timeout` and incremental SHA-256). The original function remains as the composition of the two, so the mirror path is unchanged.
- `download_blob` answers as soon as upstream has answered: `200`, `docker-content-digest`, `Content-Length` from upstream when it sent one (chunked otherwise). The body follows the spool file through a second read handle, paced by a `watch` channel of bytes flushed / done, and goes through the existing `VerifyingReader` so the streaming serve path keeps its digest check at EOF.
- The spool and `put_from_path` run in a detached task. A client that disconnects mid-fill no longer aborts the fill; the blob is cached and the client's retry is a hit. The follower's own file handle survives the storage move or the guard's unlink.
- The temp file is created by the handler (`SpoolFile`) before upstream is opened, so the follower has its handle before the first byte lands and no local failure after a circuit-breaker probe is taken can leave it unrecorded; each chunk is flushed before the progress counter advances, so a follower never reads past what is on disk.
- The follower withholds the last flushed byte until the spool reports a verified end. With a `Content-Length` body hyper stops reading once the declared bytes are out, so the digest verdict has to gate the final byte rather than EOF. A zero `Content-Length` is never declared for the same reason (an empty body goes chunked). The verdict is published right after the SHA-256 check, before `put_from_path`, so the response does not wait for the storage upload.
- A proxied digest is recorded in quarantine only once its fetch has verified. The gate still decides immediately, treating an unseen digest as freshly `Pending`, but a failed or poisoned fetch no longer starts the cooldown clock for the real blob.

### Behaviour changes

- A digest mismatch from upstream was a `502` with no bytes sent. The digest is now only known at the end of the spool, so a mismatch is a `200` whose body aborts before a clean end (the last byte is never released); the poisoned bytes are still never stored and the temp file is still cleaned up. Registry clients verify every blob's digest themselves, and this is the same failure mode the cached serve path already has through `VerifyingReader`. `test_docker_proxy_blob_sha256_mismatch_rejected` is updated accordingly.
- The blob lands in storage just after the response body completes rather than before it starts; `test_docker_proxy_blob_sha256_match_served_and_cached` now waits for it.
- Quarantine first-seen for a proxied blob is the moment its fetch verified, not the moment it was requested.

## Tests

- `test_docker_proxy_blob_streams_while_spooling`: a trickling upstream; headers arrive before the body is complete, the body matches, the blob is cached afterwards, no temp file left.
- `test_docker_proxy_blob_fill_survives_client_disconnect`: the response is dropped immediately; the blob is still cached.
- `test_docker_proxy_blob_mismatch_aborts_framed_body`, `test_docker_proxy_blob_empty_poisoned_body_aborts`: over a real listener and `reqwest`, a poisoned body (non-empty and empty) never completes and is never cached.
- `test_docker_proxy_blob_chunked_upstream_streams_and_caches`: an upstream without `Content-Length`.
- `spool_follower_*`: the follower withholds the last byte while unverified, releases it on `Ok`, errors on `Err`, and picks up bytes published between a short read and its wait.
- `test_docker_proxy_blob_sha256_mismatch_rejected`: updated to the new semantics.

`cargo fmt --check`, `cargo clippy --package nora-registry --all-targets -- -D warnings`, `cargo test --lib --bin nora` (1830 passed) are green locally.

Not addressed here, pre-existing: the upstream open path can still abandon a circuit-breaker probe on a token-fetch failure or a local write error; that predates this change.

## End-to-end validation

Alibaba Cloud ACK, NORA in `NORA_STORAGE_MODE=s3` (OSS), upstream `asia-docker.pkg.dev` over a cross-border link, probe from an in-cluster pod with `wget -S`, timestamps relative to request start:

| | status line | body complete |
|---|---|---|
| 1.0.1, cold 546 MB layer | 43 s | 49 s |
| this build, cold 1.37 GB layer | 1 s | 74 s |
| this build, same layer cached in OSS | 0 s | 14 s |

The 1.37 GB blob was in the bucket 8 s after the cold spool finished, and the cached fetch made no upstream request. A second request that arrived 3 s after the first spool finished, before the OSS upload completed, went to the other replica and fetched upstream again; that is the existing absence of cross-request coalescing on the blob path, unchanged here.

Commits are unsigned; I do not have a signing key set up on this machine.


synchrone:
The Security job is red on `cargo audit` for RUSTSEC-2026-0285 (rustls). This PR does not change `Cargo.lock`; the advisory was published after main's last CI run and #1000 already carries the rustls 0.23.43 → 0.23.45 bump. I will merge main into this branch once #1000 lands rather than duplicate the lockfile change here. Every other check is green.

synchrone:
Production data point, in case it helps review: this branch (fork build `sha-8cbab39`) has been the cache for an Alibaba ACK cluster with containerd 2.3.4 nodes since 2026-09-17 16:44 UTC. Two production rollouts since then each pulled three cold layers over 1 GB through it (largest 1.27 GB) and completed in about 8 minutes; the same rollouts had failed on every attempt since 2026-09-10 with the 1.0.1 image, one of those layers alone being retried 56 times without ever caching. All three layers now sit in the S3 (OSS) backend, later pulls are hits, and the logs show no errors over the first 24 hours. Blobs cached by 1.0.1 before the upgrade are still served.

devitway:
Root cause verified against containerd source: `ResponseHeaderTimeout: 30 * time.Second` in `DefaultHTTPTransport` (`core/remotes/docker/registry.go:267`), applied to the registry transport at `:180`, on `release/2.3` and `main`; absent on `release/2.1`/`v2.1.9`. Matches the report.

Detached fill is the right call — surviving client disconnect is what actually closes the re-fetch loop; a tee bound to the request future would still lose the fill on containerd's pull-progress cancel. Notes before merge:

1. `spawn_blob_spool` (`docker.rs:2996`) is a bare `tokio::spawn`. The other detached tasks wrap the future in `AssertUnwindSafe(...).catch_unwind()` (`main.rs:296` `spawn_cache`, `:315`, `pypi.rs:220`, `cleanup.rs:68`). As written, a panic in the fill only reaches the follower as a dropped-`watch` "abandoned" read error, with no `done=Err` and no metric. Suggest matching the existing pattern.

2. Quarantine first-seen now lands in `quarantine_record_verified` post-SHA instead of at request time on the proxy path. Gate order is fine — `quarantine_proxy_fetch_gate` (`:1376`) runs before the stream is built (`:1385`), so the GHSA-4j4m-fchf-gr9r hold still applies — but the first-seen clock shifts by the fetch duration. Worth an explicit test that a held new digest is still 403 on the streaming path.

3. 502→200 on upstream digest mismatch: inherent once headers ship, and it matches the local-hit `VerifyingReader` failure mode, so acceptable — but it's a behavior change that deserves a maintainer ack, not just a test update. You already flagged it in #1001.

Per-request spool doesn't single-flight: N concurrent cold pulls of one digest = N upstream fetches + N `put_from_path` on one key (safe via atomic store, N× egress). That's the blob-coalesce gap #595 left open, not this PR's job — filed as #1003.

Disclosed already, non-blocking: CB probe dropped on token-fetch failure (`open_blob_from_upstream:3125`); temp-file create failure → 500 with no upstream degrade (`:1328`).

synchrone:
Thanks for the review. Both notes addressed in 218fcf0, rebased on your merge of main:

1. `spawn_blob_spool` now runs the fill under `AssertUnwindSafe(..).catch_unwind()` like `spawn_cache` and the PyPI backfill. A panic is logged, counted as `nora_cache_write_errors_total{registry="docker",operation="proxy_blob_spool_panic"}`, and published to the follower as `done = Err`, so the response aborts on a failed spool rather than on a dropped sender. A failed `put_from_path` is counted under `operation="proxy_blob_store"` as well.
2. `test_docker_proxy_blob_new_digest_held_on_streaming_path`: with quarantine in enforce mode, a never-seen digest gets 403 before any byte on the streaming path, the fill still lands in the cache, `digest_store.check` reports it Pending once the fetch verified, and a second request is still held.
3. The 502 → 200-with-aborted-body change on upstream digest mismatch stays as is; agreed it is inherent once headers ship, and it is yours to ack.

The first-seen clock moving from request time to verification time is deliberate: it means a fetch that fails or is poisoned never starts the cooldown for the real blob. Happy to revert to request-time recording if you would rather keep the old semantics.

#1003 for the coalescing gap makes sense; the production numbers in my earlier comment show exactly that pattern (a second replica re-fetching a blob whose upload was still in flight).

ertime037 (approved):

Issue #374, opened 2026-05-18
Refactor: decompose docker.rs (21 changes in 50 commits)


devitway:
## Problem

`docker.rs` changed 21 times in the last 50 commits — the highest churn file in the codebase. High churn suggests it handles too many concerns (god object).

**Other high-churn files:**
- `main.rs` — 14 changes
- `pypi.rs` — 13 changes
- `maven.rs` — 13 changes
- `cargo_registry.rs` — 12 changes
- `config.rs` — 11 changes

## Proposal

Split `docker.rs` into focused modules:

```
registry/docker/
├── mod.rs          — route registration, public API
├── proxy.rs        — upstream proxy logic, namespace resolution
├── upload.rs       — chunked upload state machine
├── manifest.rs     — manifest get/put/delete
├── catalog.rs      — catalog and tag listing
└── auth.rs         — (already extracted as docker_auth.rs)
```

This reduces merge conflicts, makes code review easier, and isolates blast radius of changes.

## Criteria

- No behavior change — pure structural refactor
- All existing tests pass
- Each sub-module < 500 lines

PR #1004, opened 2026-09-19
feat(cpan): add CPAN proxy registry

0x62ash:
## What

Adds a CPAN proxy registry with cached package indexes and immutable distribution archives, curation and observability integration, OpenAPI documentation, and dashboard/UI support.

Cached archives are grouped by distribution on author pages. The canonical distribution page lists all cached versions, redirects old archive-shaped UI URLs, and shows `cpanm --from {NORA_URL}/cpan Package::Name`. A Playwright scenario covers the complete flow.

## Why

NORA did not support Perl/CPAN clients, and cached releases had no distribution-level UI where users could see every available version.

## Validation

- `cargo fmt --check`
- `cargo clippy --package nora-registry -- -D warnings`
- `cargo test --lib --bin nora` (63 library and 1833 binary tests passed; 1 ignored)
- `npx playwright test tests/cpan-proxy.spec.ts` (3 passed against an isolated local CPAN upstream)
- `./scripts/changelog-fragments.sh --check`

## Checklist

- [x] Added `changelog.d/1004.added.md`
- [x] New registry checklist completed


devitway:
Built this branch and ran a live smoke against a real CPAN mirror (`cpan.metacpan.org` — identical content to `www.cpan.org`, which resolved IPv6-only from where I tested):

- `GET /cpan/modules/02packages.details.txt.gz` proxied the real 2.5 MB index (259,897 module lines), parsed and cached; a second fetch was a ~14× faster cache hit.
- Distributions served as valid gzip, including a 17.9 MB one and a `v`-string version (`Crypt-HSXKPasswd-v3.6`) — no trouble on the messier real names.
- The hierarchical `/ui/cpan` browse rendered on the cached data; the homepage shows the CPAN card + icon.
- Zero panics or errors across the whole run.

So the parser/index and the browse hold up on real upstream data, not just the fixture. Trust-boundary wiring reads correctly too: the proxy path gates through `digest_quarantine::proxy_gate`/`proxy_gate_dated`, `curation::check_download` and `check_namespace_isolation`, and `RegistryType::Cpan` extends the enum so `/health` and metrics pick it up via `all()`/`enabled_registries` (no per-file changes needed there).

Non-blocking note: `cpan.rs` is a large single-file handler, same shape as `docker.rs` — fine alongside the other per-format handlers; decomposition is its own track (cf. #374).

Correctness-wise this looks good to me. Pending the usual before merge: rebase onto `main` (currently behind) and a green CI run.

ertime037 (approved):

devitway (approved):

